PR #7, opened 2015-06-06
More robust canAccessKeychain

dfed:
Looks like you can write to the keychain on OS X even when you don't have read access (or at least find out that you're trying to add a duplicate item when you don't have access). So instead of cleverly trying to discern the error message behind a SecItemAdd, add the item to the keychain and then see if we can read it back out.

Also did some cleanup to allow VALSecureEnclaveValet to be unit tested.

### DIFF
--- a/Valet/VALSecureEnclaveValet.m
+++ b/Valet/VALSecureEnclaveValet.m
@@ -32,7 +32,7 @@
 {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wtautological-compare"
-#if TARGET_OS_IPHONE && __IPHONE_8_0
+#if VAL_IOS_8_OR_LATER
     return (&kSecAttrAccessControl != NULL && &kSecUseOperationPrompt != NULL);
 #else
     return NO;
@@ -85,7 +85,7 @@
 
 - (BOOL)containsObjectForKey:(NSString *)key;
 {
-#if TARGET_OS_IPHONE && __IPHONE_8_0
+#if VAL_IOS_8_OR_LATER
     OSStatus status = [self containsObjectForKey:key options:@{ (__bridge id)kSecUseNoAuthenticationUI : @YES }];
     BOOL const keyAlreadyInKeychain = (status == errSecInteractionNotAllowed);
     return keyAlreadyInKeychain;
@@ -101,7 +101,7 @@
 
 - (NSError *)migrateObjectsMatchingQuery:(NSDictionary *)secItemQuery removeOnCompletion:(BOOL)remove;
 {
-#if TARGET_OS_IPHONE && __IPHONE_8_0
+#if VAL_IOS_8_OR_LATER
     if ([[self class] supportsSecureEnclaveKeychainItems]) {
         VALCheckCondition(secItemQuery[(__bridge id)kSecUseOperationPrompt] == nil, [NSError errorWithDomain:VALMigrationErrorDomain code:VALMigrationErrorInvalidQuery userInfo:nil], @"kSecUseOperationPrompt is not supported in a migration query. Keychain items can not be migrated en masse from the Secure Enclave.");
     }
@@ -114,7 +114,7 @@
 
 - (BOOL)setObject:(NSData *)value forKey:(NSString *)key userPrompt:(NSString *)userPrompt
 {
-#if TARGET_OS_IPHONE && __IPHONE_8_0
+#if VAL_IOS_8_OR_LATER
     return [self setObject:value forKey:key options:@{ (__bridge id)kSecUseOperationPrompt : userPrompt }];
 #else
     return NO;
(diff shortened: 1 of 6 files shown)


dfed:
cc @EricMuller22 @mikelikespie @mthole

Valet/VALValet.m
```diff
@@ -227,15 +227,25 @@ - (BOOL)isEqualToValet:(VALValet *)otherValet;
...
 - (BOOL)canAccessKeychain;
 {
-    NSString *canaryKey = @"VAL_KeychainCanaryUsername";
-    NSString *canaryValue = @"VAL_KeychainCanaryPassword";
+    BOOL canAccessKeychain = NO;
     
-    NSMutableDictionary *query = [self.baseQuery mutableCopy];
-    [query addEntriesFromDictionary:[self _secItemFormatDictionaryWithKey:canaryKey]];
-    [query addEntriesFromDictionary:@{ (__bridge id)kSecValueData : [canaryValue dataUsingEncoding:NSUTF8StringEncoding] }];
+    [self.lockForSetAndRemoveOperations lock];
```
mikelikespie: Probably should be in a `@try`/`@finally` to make sure its unlocked if an exception is thrown.

ValetTests/ValetTests.m
```diff
@@ -64,14 +67,21 @@ - (void)setUp;
...
     [super setUp];
     
-    self.valet = [[VALValet alloc] initWithIdentifier:@"valet_testing" accessibility:VALAccessibilityWhenUnlocked];
-    self.testingValet = [[VALTestingValet alloc] initWithIdentifier:@"valet_testing" accessibility:VALAccessibilityWhenUnlocked];
-    self.synchronizableValet = [[VALSynchronizableValet alloc] initWithIdentifier:@"valet_testing" accessibility:VALAccessibilityWhenUnlocked];
+    NSString *const valetTestingIdentifier = @"valet_testing";
+    self.valet = [[VALValet alloc] initWithIdentifier:valetTestingIdentifier accessibility:VALAccessibilityWhenUnlocked];
+    self.testingValet = [[VALTestingValet alloc] initWithIdentifier:valetTestingIdentifier accessibility:VALAccessibilityWhenUnlocked];
+    self.synchronizableValet = [[VALSynchronizableValet alloc] initWithIdentifier:valetTestingIdentifier accessibility:VALAccessibilityWhenUnlocked];
+#if TARGET_OS_IPHONE && __IPHONE_8_0
```
mikelikespie: Would be nice to see some comments on why these macros.  also possbly maybe do a `#define VAL_IS_IPHONE_8_OR_GREATER (TARGET_OS_IPHONE && __IPHONE_8_0)` somewhere to make it so you're not repeating `TARGET_OS_IPHONE && __IPHONE_8_0` all over.

mikelikespie:
LGTM otherwise

Valet/VALValet.m
```diff
@@ -227,15 +227,25 @@ - (BOOL)isEqualToValet:(VALValet *)otherValet;
...
+    {
+        NSString *const canaryKey = @"VAL_KeychainCanaryUsername";
+        NSString *const canaryValue = @"VAL_KeychainCanaryPassword";
+        
+        // Manually add the key to the keychain since we don't care about duplicates and are optimizing for speed.
+        NSMutableDictionary *query = [self.baseQuery mutableCopy];
+        [query addEntriesFromDictionary:[self _secItemFormatDictionaryWithKey:canaryKey]];
+        [query addEntriesFromDictionary:@{ (__bridge id)kSecValueData : [canaryValue dataUsingEncoding:NSUTF8StringEncoding] }];
```
mikelikespie: If we're adding one key/value, may be more efficient to do `-[setValue:forKey:]`

ValetTests/ValetTests.m
dfed: solid call.

Valet/VALValet.m
dfed: I'd done addEntriesFromDictionary for legibility... but subscripting syntax is also pretty darned legible. will change.

Valet/VALSecureEnclaveValet.m
```diff
@@ -32,7 +32,7 @@ + (BOOL)supportsSecureEnclaveKeychainItems;
 {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wtautological-compare"
-#if TARGET_OS_IPHONE && __IPHONE_8_0 && !TARGET_IPHONE_SIMULATOR
+#if VAL_IOS_8_OR_LATER
```
dfed: Surprisingly, Secure Enclave storage works in the simulator.

Valet/VALValet.m
mikelikespie: Oh yeah, forgot about subscripting syntax. That's the right call.

mthole:
:+1:

PR #9, opened 2015-06-19
Vulnerability test should not rely on Finder

dfed:
SystemUIServer has lived in a stable location for many OS releases, and it is impossible to run your Mac without it.

cc @EricMuller22 @martinmroz @kyleve


kyleve:
LGTM

EricMuller22:
Gorm

dfed:
Travis-CI's Mac builds are currently down... waiting for a green build before merging.

https://www.traviscistatus.com

dfed:
Not bumping the minimum version target to 10.10 because Travis CI can't handle it yet. Unfortunate, but it is what it is. Merging the rest as-is.

Issue #10, opened 2015-06-22
Debug-only read errors: errSecMissingEntitlement -34018


jcampbell05:
Hey, Do you know if its possible to use the same instance across threads ?

We have an issue where we can write but we don't get the data back when reading. However when we run this on the main queue it is okay.

dfed:
Valet is built with the intent to allow allow using the same instance across multiple threads, so this should indeed work. Is it possible you've got a timing issue? Try adding a break point on the return line of [objectForKey:](https://github.com/square/Valet/blob/master/Valet/VALValet.m#L475) and [setObject:forKey:](https://github.com/square/Valet/blob/master/Valet/VALValet.m#L457) to see what your timing is. If the order of operations is what you're expecting, I'm curious what the status code is on the failed `objectForKey:`

EricMuller22:
And just to help with diagnosis, mind providing us with which platform and which Valet subclass you're using?

jcampbell05:
This is as much as we know.

We added some breakpoints and they execute in the correct order, we tried this using `FXKeychain` and it logs out an error code -34018 (Something I think Valet should do). 

This error seems to mean that either our bundle ID doesn't match the one used to write/read to the keychain or we don't have the correct entitlements if it's a shared one.

We used the normal Valet subclass for accessing the keychain, the weird thing is when we run the exact same app not from Xcode it works perfectly. And sometimes running it from Xcode it will be perfectly okay especially if you've ran the app untethered beforehand but then suddenly we will just get that error.

We've tried to research this online, it seems some people have this issue but so far no solutions.

dfed:
This is a debug-only entitlement issue, not a threading issue. Sometimes Xcode fails to sign the debug build properly, which can cause this error. We saw this a whole lot on Xcode 6.2 IIRC, but we haven't seen it since upgrading to 6.3. `[valet canAccessKeychain]` will return NO in this state. We worked around this issue by asserting `canAccessKeychain` in `application:didFinishLaunchingWithOptions:` under `#if DEBUG`. That way we knew right off the bat if Xcode/entitlements had screwed up. This issue *never* affected our signed enterprise builds or our signed app store builds.

Any chance you aren't running the latest version of Xcode?

jcampbell05:
So we did some more research and we found out that one of our developers is using OS X 10.11. And it seems the combination of Xcode 6.3.2, OS X 10.11 and iOS 8.3 causes this issue.

[...]

PR #11, opened 2015-06-22
Add tests for writing and then reading from different threads with expected results.

dfed:
cc @EricMuller22 @pwesten 

Explicitly tests the issue brought up by https://github.com/square/Valet/issues/10

### DIFF
--- a/ValetTests/ValetTests.m
+++ b/ValetTests/ValetTests.m
@@ -281,6 +281,49 @@
     [self waitForExpectationsWithTimeout:5.0 handler:nil];
 }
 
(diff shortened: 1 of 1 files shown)


EricMuller22:
LGTM, let's wait and see what we hear about which Valet (and which platform) the issue may be occurring on though.

dfed:
Merging this since there's nothing destructive here. @jcampbell05, I'm curious how your setup differs from the unit test. Let us know! It's plausible there's some keychain configuration oddness out there that we need to protect against.

PR #12, opened 2015-06-24
Add debugging tips to the README

dfed:
cc @EricMuller22 @mikelikespie 

I believe this enumerates all failure cases. Let me know if you can think of others.


EricMuller22:
Gorm.

PR #15, opened 2015-06-30
Use Xcode 6.3 beta on TravisCI

dfed:
Trying out Xcode 6.3 CI. Should allow us to use `nonnull` and `nullable` syntax instead of `__attribute` syntax.

cc @EricMuller22 @mtauraso @rhgills

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: objective-c
+osx_image: beta-xcode6.3
 script:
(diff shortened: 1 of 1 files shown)


dfed:
Ignore the CircleCI failure. Currently looking into how to turn that off.

dfed:
Closed this because Xcode 6.4 is out on Travis :tada:

PR #25, opened 2015-07-14
Fix typo in README

ianyh:
Changed `writen` to `written`.


dfed:
Good catch! Thanks :)

PR #27, opened 2015-07-23
Valet to 2.0

dfed:
• Support Xcode 7
• Treat warnings as errors in all targets
• Breaking change: Remove `set*:forKey:prompt:` on VALSecureEnclaveValet since `prompt` will never be used.
• Do not use `NS_ASSUME_NONNULL_*`. Explicitly use `nullable` or `nonnull` to avoid oversight.
• Fix memory leak
• Update to recommended build settings on Xcode 7 (everything still works on Xcode 6.4)

Given the breaking change and support for a brand new Xcode, it made sense to go to v2.0.

cc @EricMuller22 @pwesten @kenwigginton @mthole

### DIFF
--- a/Valet.podspec
+++ b/Valet.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'Valet'
-  s.version  = '1.3.1'
+  s.version  = '2.0'
   s.license  = 'Apache License, Version 2.0'
   s.summary  = 'Valet lets you securely store data in the iOS or OS X Keychain without knowing a thing about how the Keychain works. It\'s easy. We promise.'
   s.homepage = 'https://github.com/square/Valet'
--- a/Valet/VALSecureEnclaveValet.h
+++ b/Valet/VALSecureEnclaveValet.h
@@ -21,9 +21,6 @@
 #import "VALValet.h"
 
 
-NS_ASSUME_NONNULL_BEGIN
-
-
 /// Reads and writes keychain elements that are stored on the Secure Enclave (supported on iOS 8.0 or later) using accessibility attribute VALAccessibilityWhenPasscodeSetThisDeviceOnly. Accessing or modifying these items will require the user to confirm their presence via Touch ID or passcode entry. If no passcode is set on the device, the below methods will fail. Data is removed from the Secure Enclave when the user removes a passcode from the device. Use the userPrompt methods to display custom text to the user in Apple's Touch ID and passcode entry UI.
 NS_CLASS_AVAILABLE_IOS(8_0)
 @interface VALSecureEnclaveValet : VALValet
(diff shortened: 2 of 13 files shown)


Valet.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'Valet'
-  s.version  = '1.3.1'
+  s.version  = '2.0'
```
kenwigginton: :+1:

EricMuller22:
Just talked in person - looks good overall (the memory leak was a nice catch and I'm all about warnings as errors). I balked initially at taking out the `NS_ASSUME_NONNULL`s, but I buy that being more explicit should help avoid bugs. Going to take a closer look at that commit then dole out the gorm.

Valet/VALSecureEnclaveValet.h
```diff
@@ -32,33 +29,24 @@ NS_CLASS_AVAILABLE_IOS(8_0)
...
 
 /// Creates a Valet that reads/writes Secure Enclave keychain elements that can be shared across applications written by the same development team.
 /// @param sharedAccessGroupIdentifier This must correspond with the value for keychain-access-groups in your Entitlements file.
-- (nullable instancetype)initWithSharedAccessGroupIdentifier:(NSString *)sharedAccessGroupIdentifier;
+- (nullable instancetype)initWithSharedAccessGroupIdentifier:(nonnull NSString *)sharedAccessGroupIdentifier;
 
-/// Convenience method for inserting data into the keychain with a user prompt.
-/// @param userPrompt The prompt displayed to the user in Apple's Touch ID and passcode entry UI when updating a value.
-- (BOOL)setObject:(NSData *)value forKey:(NSString *)key userPrompt:(NSString *)userPrompt;
 /// Convenience method for retrieving data from the keychain with a user prompt.
```
NickEntin: Might be good to specify here when the method would return null.  Unclear if it's when the key doesn't have a value, when permissions are denied, etc.  Same for the string version below.

dfed: You're right that the README is a lot clearer on when this will return `nil` vs the docs here. I can update.

Summary: `objectForKey:` and `stringForKey` return `nil` whenever `canAccessKeychain` returns `NO`, or when the object doesn't exist in the keychain.

Valet/VALValet.m
```diff
@@ -167,7 +167,12 @@ + (VALValet *)_sharedValetForValet:(VALValet *)valet;
 
 #pragma mark - Initialization
 
-- (instancetype)initWithIdentifier:(NSString *)identifier accessibility:(VALAccessibility)accessibility;
+- (nullable instancetype)init NS_UNAVAILABLE;
+{
+    VALCheckCondition(NO, nil, @"Use a designated initializer");
+}
+
+- (nullable instancetype)initWithIdentifier:(NSString *)identifier accessibility:(VALAccessibility)accessibility;
```
NickEntin: Specify `(nonnull NSString *)identifier`

Valet/VALValet.m
```diff
@@ -184,7 +189,7 @@ - (instancetype)initWithIdentifier:(NSString *)identifier accessibility:(VALAcce
     return [[self class] _sharedValetForValet:self];
 }
 
-- (instancetype)initWithSharedAccessGroupIdentifier:(NSString *)sharedAccessGroupIdentifier accessibility:(VALAccessibility)accessibility;
+- (nullable instancetype)initWithSharedAccessGroupIdentifier:(NSString *)sharedAccessGroupIdentifier accessibility:(VALAccessibility)accessibility;
```
NickEntin: Specify `(nonnull NSString *)sharedAccessGroupIdentifier`

Valet/VALValet.m
```diff
@@ -224,7 +229,7 @@ - (NSString *)description;
 
 #pragma mark - NSCopying
 
-- (instancetype)copyWithZone:(NSZone *)zone;
+- (instancetype)copyWithZone:(nullable NSZone *)zone;
```
NickEntin: Specify `nonnull instancetype`?  I don't think `self` can be `nil`...

Valet/VALValet.h
```diff
@@ -71,19 +68,19 @@ typedef NS_ENUM(NSUInteger, VALMigrationError) {
...
+- (nullable instancetype)initWithSharedAccessGroupIdentifier:(nonnull NSString *)sharedAccessGroupIdentifier accessibility:(VALAccessibility)accessibility NS_DESIGNATED_INITIALIZER;
 
-@property (copy, readonly) NSString *identifier;
+@property (nonnull, copy, readonly) NSString *identifier;
 @property (readonly, getter=isSharedAcrossApplications) BOOL sharedAcrossApplications;
 @property (readonly) VALAccessibility accessibility;
 
 /// @return YES if otherValet reads from and writes to the same sandbox within the keychain as the receiver.
-- (BOOL)isEqualToValet:(VALValet *)otherValet;
+- (BOOL)isEqualToValet:(nonnull VALValet *)otherValet;
```
NickEntin: Specify `nonnull` for this method in .m file

Valet/VALValet.m
dfed: indeed it must be nonnull per the Xcode 7 SDK.

Valet/VALValet.m
dfed: good catch! annoying that the compiler doesn't consider this an error. I guess if the .h declares it the .m doesn't need it... but that sucks.

NickEntin: Or at least a warning.  Worth a radar?

dfed: Yeah, probably. I can write it up later today.

Valet/VALValet.h
dfed: done!

Valet/VALSecureEnclaveValet.h
dfed: This is now fixed. Let me know if you like it!

NickEntin:
LGTM   Good job as always @dfed!

dfed:
Holding merging until I get a GORM from @EricMuller22. Of course, all eyes are welcome!

EricMuller22:
ᗧ······ᗣ

(gorm)

EricMuller22:
We are probably going to want to do a subsequent pass and make sure we've got (tests and) proper check conditions in place for the nonnull parameters

dfed:
@EricMuller22 just did a pass. While we don't necessarily have tests for all these things, the check conditions do exist. Merging as-is.

PR #28, opened 2015-08-02
Expose unavailable initializer in header

dfed:
cc @kyleve @EricMuller22 

LMK if you think this is worthy of a minor version bump


dfed:
gorm, probably worth bumping though

EricMuller22:
gorm, probably worth bumping though

dfed:
Bumped. Will merge when green

PR #36, opened 2015-08-25
Support use of kSecUseNoAuthenticationUI on iOS 8 and Xcode 7

dfed:
cc @EricMuller22 @Visorox

Bit of a facepalm... was too fast on the previous PR. Plan is to replace the existing pod version so 2.0.2 doesn't crash on Xcode 7 and iOS 8.


EricMuller22:
Looks gorm

Issue #35, opened 2015-08-25
Carthage support


Visorox:
So as some of you may know [Carthage](https://github.com/Carthage/Carthage) is becoming a popular tool for handling dependencies. Carthage is currently iOS 8+ exclusive due to it using frameworks.

What are your thoughts on supporting this?

My basic understanding of how this would work is by duplicating the current static library targets.

dfed:
If supporting Carthage does not affect Cocoapods consumers (or consumers who are still targeting iOS 7), I see no harm in supporting it.

Is there any kind of changes we should make to our CI to support Carthage (or prevent breaking Carthage), in addition to the target changes?

dfed:
Once #42 is in I want to add a Carthage gem to the README

dfed:
@Visorox @eliperkins @dcaunt I've uploaded the Carthage binary to Github. Let me know if it works!

eliperkins:
![screen shot 2015-09-16 at 6 36 59 pm](https://cloud.githubusercontent.com/assets/1051453/9920433/02aced64-5ca2-11e5-9b08-2b5c63716479.png)

https://github.com/eliperkins/ValetExample/commit/e2e4b8d3f01a4f4d33913866e9d691a40efe8214

Works for me! Great job, y'all!

dcaunt:
Looks good to me! 

By the way, if you give your Releases a name (e.g. '2.0.2' rather than using the latest commit message, you'll get more useful output from Carthage. This can be helpful when you've specified a flexible version.

dfed:
@dcaunt didn't realize Carthage didn't version based on git tags. Thanks for raising! This should now be fixed.

PR #37, opened 2015-08-25
Framework/Carthage support

EricMuller22:
In-flight/WIP branch to add Carthage support #35 

cc @dfed

### DIFF
--- a/Valet.xcodeproj/project.pbxproj
+++ b/Valet.xcodeproj/project.pbxproj
@@ -7,6 +7,9 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		AC2C2D481B8CE184008F4EE8 /* Valet_iOS_Framework.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = AC2C2D311B8CE184008F4EE8 /* Valet_iOS_Framework.framework */; };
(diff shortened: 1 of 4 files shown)


Valet.xcodeproj/project.pbxproj
```diff
@@ -7,6 +7,9 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		AC2C2D481B8CE184008F4EE8 /* Valet_iOS_Framework.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = AC2C2D311B8CE184008F4EE8 /* Valet_iOS_Framework.framework */; };
```
dfed: do we want an OS X framework as well? Or does Carthage not do OS X (#newb)?

EricMuller22:
Just started with iOS, set the deployment target to 8.0 - build output here:
screen shot 2015-08-25 at 11 01 21 am

dfed:
Let's make sure we don't break module maps. Cc @mikelikespie, our resident expert on them.

Valet.xcodeproj/project.pbxproj
eliperkins: What do you think about making `Valet.framework` should be the output name of the framework here? This file will be nested inside of `Carthage/Builds/iOS` anyway, so the additional naming of iOS is redundant, and `_Framework.framework` is also redundant.

eliperkins: Also, Carthage does OS X!

EricMuller22:
Closing in favor of #42. Thanks @eliperkins!

PR #42, opened 2015-09-15
Add framework targets for iOS and OS X

### LINKED ISSUE #35: Carthage support

eliperkins:
Closes #35 

This will allow the project to be used by Carthage and other dynamic framework tooling.

Based loosely off of #37 from @EricMuller22, with a couple of changes. This adds two new targets for iOS and OS X that are both dynamic frameworks, compiling the source on their own, rather than including the static library into the dynamic framework.

This also scopes out the headers into their correct visibility, so we shouldn't ever get to see anything we shouldn't, like `mutableBaseQueryWithIdentifier:initializer:accessibility:` or any of it's friends.

Lastly, I renamed the framework files to be `Valet.framework` to better support universal frameworks.

### DIFF
--- a/Valet.xcodeproj/project.pbxproj
+++ b/Valet.xcodeproj/project.pbxproj
@@ -17,8 +17,8 @@
 		26F06A591BA8B55100E039CD /* VALValet_Protected.h in Headers */ = {isa = PBXBuildFile; fileRef = EAEEAC231AB7BA0C00EDB6E3 /* VALValet_Protected.h */; };
 		26F06A5A1BA8B55300E039CD /* VALSecureEnclaveValet.h in Headers */ = {isa = PBXBuildFile; fileRef = EAEEAC1E1AB7B84E00EDB6E3 /* VALSecureEnclaveValet.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		26F06A5B1BA8B55500E039CD /* VALSynchronizableValet.h in Headers */ = {isa = PBXBuildFile; fileRef = EAEEAC1B1AB7B84000EDB6E3 /* VALSynchronizableValet.h */; settings = {ATTRIBUTES = (Public, ); }; };
-		26F06A5C1BA8B55D00E039CD /* ValetDefines.h in Headers */ = {isa = PBXBuildFile; fileRef = EAEEAC251AB7BA9800EDB6E3 /* ValetDefines.h */; settings = {ATTRIBUTES = (Public, ); }; };
-		26F06A5D1BA8B56000E039CD /* ValetDefines.h in Headers */ = {isa = PBXBuildFile; fileRef = EAEEAC251AB7BA9800EDB6E3 /* ValetDefines.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		26F06A5C1BA8B55D00E039CD /* ValetDefines.h in Headers */ = {isa = PBXBuildFile; fileRef = EAEEAC251AB7BA9800EDB6E3 /* ValetDefines.h */; };
+		26F06A5D1BA8B56000E039CD /* ValetDefines.h in Headers */ = {isa = PBXBuildFile; fileRef = EAEEAC251AB7BA9800EDB6E3 /* ValetDefines.h */; };
 		EA1E1F8F1A8C46090067C991 /* libValet.a in Frameworks */ = {isa = PBXBuildFile; fileRef = EA1E1F831A8C46080067C991 /* libValet.a */; };
 		EA1E1FA01A8C48560067C991 /* ValetTests.m in Sources */ = {isa = PBXBuildFile; fileRef = EA1E1F9F1A8C48560067C991 /* ValetTests.m */; };
 		EAEAA88D1B167A8700F7AA98 /* libValet.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = EAEAA8821B167A8600F7AA98 /* libValet.dylib */; };
@@ -273,8 +273,8 @@
 				26F06A581BA8B55000E039CD /* VALValet.h in Headers */,
 				26F06A5A1BA8B55300E039CD /* VALSecureEnclaveValet.h in Headers */,
 				26F06A5B1BA8B55500E039CD /* VALSynchronizableValet.h in Headers */,
-				26F06A5D1BA8B56000E039CD /* ValetDefines.h in Headers */,
 				26F06A591BA8B55100E039CD /* VALValet_Protected.h in Headers */,
+				26F06A5D1BA8B56000E039CD /* ValetDefines.h in Headers */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
--- a/Valet.xcodeproj/project.pbxproj
+++ b/Valet.xcodeproj/project.pbxproj
@@ -652,7 +652,7 @@
 				GCC_NO_COMMON_BLOCKS = YES;
 				INFOPLIST_FILE = Valet/Info.plist;
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
-				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = com.square.Valet;
 				PRODUCT_NAME = Valet;
@@ -677,7 +677,7 @@
 				GCC_NO_COMMON_BLOCKS = YES;
 				INFOPLIST_FILE = Valet/Info.plist;
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
-				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = com.square.Valet;
 				PRODUCT_NAME = Valet;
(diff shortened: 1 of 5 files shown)


dfed:
Thanks for the PR! To contribute to Valet you must first sign our [CLA](https://docs.google.com/a/squareup.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1).

Taking a look at the diff now.

Valet.xcodeproj/project.pbxproj
```diff
@@ -497,6 +639,106 @@
...
+				CURRENT_PROJECT_VERSION = 1;
+				DEBUG_INFORMATION_FORMAT = dwarf;
+				DEFINES_MODULE = YES;
+				DYLIB_COMPATIBILITY_VERSION = 1;
+				DYLIB_CURRENT_VERSION = 1;
+				DYLIB_INSTALL_NAME_BASE = "@rpath";
+				GCC_NO_COMMON_BLOCKS = YES;
+				INFOPLIST_FILE = Valet/Info.plist;
+				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
+				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
```
eliperkins: This can probably get throttled back to `6.0`.

eliperkins:
@dfed Ah, sorry about that! Signed.

Valet.xcodeproj/project.pbxproj
dfed: Please do!

eliperkins:
Ah, meant to mention I created an example repo for testing too: https://github.com/eliperkins/ValetExample

dfed:
Thanks for signing! And awesome re the ValetExample repo. Forgive me for the following questions: I'm new to Carthage.

Anything we can do in CI to ensure that we don't break Carthage? Or is our normal build/test suite enough? Also, do we need to do anything specific to update Carthage when we make new releases?

Valet.xcodeproj/project.pbxproj
```diff
@@ -497,6 +639,106 @@
 /* End PBXVariantGroup section */
 
 /* Begin XCBuildConfiguration section */
+		26E682811BA8B3F900EFF4EA /* Debug */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
```
dfed: This line seems dubious. @mtauraso, thoughts?

eliperkins:
@dfed One possibility for testing in CI might be to use the framework somewhere in testing, possibly in an existing test target or a new test target. That would at least guarantee the support of the dynamic framework as more of an integration test. Beyond that, I'm not sure about best practices akin to `pod lib lint`.

In terms of keeping Carthage happy, since Carthage relies on git and GitHub under the hood, keeping releases up to date is a great idea. Creating tags and an associated GitHub release, uploading the archive from `carthage archive Valet` is the best way to keep Carthage consumers happy.

Valet.xcodeproj/project.pbxproj
```diff
@@ -497,6 +639,106 @@
...
+				CURRENT_PROJECT_VERSION = 1;
```
dfed: @eliperkins any idea how this value is used / if it is used by Carthage?

My guess is "no" given [Apple's documentation on how to use it](https://developer.apple.com/library/ios/qa/qa1827/_index.html). But it's a little odd that the CURRENT_PROJECT_VERSION doesn't match the actual framework version.

Valet.xcodeproj/project.pbxproj
dfed: Apparently we already had this in Valet master... so likely not a problem.

Valet.xcodeproj/project.pbxproj
eliperkins: @dfed Correct, unused by Carthage. Carthage relies on git tags for versioning.

dfed: Also looks like we already had `CURRENT_PROJECT_VERSION = 1;` in our `Valet Mac` target. TIL.

Valet.xcodeproj/project.pbxproj
eliperkins: @dfed Agreed. Shouldn't be a problem though, given a peek at other libraries depending on/using Carthage (see [ReactiveCocoa](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/e16f47cf9cb568136ebd81430b24af274c3c27c7/ReactiveCocoa.xcodeproj/project.pbxproj#L2058), [Quick](https://github.com/Quick/Quick/blob/ac5eadd202dd67741628f80fb96e8c57388d954d/Quick.xcodeproj/project.pbxproj#L1168)).

Valet.xcodeproj/project.pbxproj
eliperkins: We can strip these settings out too, and fall back on Xcode's brains. These are just baseline targets that I generated by creating new Cocoa and Cocoa Touch Framework targets in Xcode. Let me know if you think we should just rip them out!

Valet/Info.plist
```diff
@@ -0,0 +1,28 @@
...
+	<key>CFBundlePackageType</key>
+	<string>FMWK</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
+	<key>CFBundleSignature</key>
+	<string>????</string>
+	<key>CFBundleVersion</key>
+	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<key>NSHumanReadableCopyright</key>
+	<string>Copyright © 2015 Square, Inc. All rights reserved.</string>
```
dfed: nit: remove `All rights reserved.` Given our license we do not reserve all rights

Valet/Info.plist
```diff
@@ -0,0 +1,28 @@
...
+	<key>NSPrincipalClass</key>
+	<string></string>
```
dfed: Should this be empty?

Valet.xcodeproj/project.pbxproj
dfed: We can leave them in, given that we already had them elsewhere. I can do a pass to remove unused key/values from the project file another time.

Valet/Info.plist
eliperkins: From the [docs](https://developer.apple.com/library/mac/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html#//apple_ref/doc/uid/TP40009251-SW11):
> NSPrincipalClass (String - OS X). This key contains a string with the name of a bundle’s principal class. This key is used to identify the entry point for dynamically loaded code, such as plug-ins and other dynamically-loaded bundles. The principal class of a bundle typically controls all other classes in the bundle and mediates between those classes and any classes outside the bundle. The class identified by this value can be retrieved using the principalClass method of NSBundle. For Cocoa apps, the value for this key is NSApplication by default.

It looks like this is pretty unnecessary for frameworks and something that is reserved more for applications. This key, incidentally, does not do anything on iOS haha.

dfed:
Good to know re github versioning and `carthage archive Valet`! Let me do some research on keeping Carthage happy to see if we can add anything to our CI.

dcaunt:
If 'carthage build --no-skip-current' completes without error, Carthage is happy. This is your pod lint equivalent, if you like.

dfed:
Awesome @dcaunt. @eliperkins let's add `- carthage build --no-skip-current --verbose` to the end of our [.travis.yml](https://github.com/square/Valet/blob/master/.travis.yml) file.

I was about to suggest we add the following lines to .travis.yml:
```
    - xcodebuild -project Valet.xcodeproj -scheme "Valet-iOS" -sdk iphonesimulator -configuration Debug -PBXBuildsContinueAfterErrors=0 ACTIVE_ARCH_ONLY=0 build
    - xcodebuild -project Valet.xcodeproj -scheme "Valet-Mac" -sdk iphonesimulator -configuration Debug -PBXBuildsContinueAfterErrors=0 ACTIVE_ARCH_ONLY=0 build
```

But it looks like `carthage build` essentially does this for us.

eliperkins:
Does Travis CI have Carthage by default? Looking through https://github.com/travis-infrastructure/osx-image-bootstrap/blob/1188d90aafa80370185f285ff1a26950d2a1a216/bootstrap.sh, I don't believe it does. Otherwise, we'll have to `brew install carthage`. 

Adding the two `xcodebuild` command could be a good idea. The unfortunate part is we can't see/test against the desired output, which would be a valid directory at `Carthage/Build/iOS/Valet.framework`. I think adding these scripts would still be beneficial though!

dfed:
My guess is we'll need to do the `brew install`.

I don't think we we need the `xcodebuild` command in addition to the `carthage build` command. If you want to build the desired output, you could run:
```
    - carthage build --no-skip-current --verbose
    - carthage archive Valet
```

I'm not sure how/if you'd want to test that the framework that's built works, though. Maybe rename `ValetExample` to `ValetCarthageTest` and check it in, then cd to ValetCarthageTest and try to link the framework? Though my (blind) guess is if the framework builds successfully it's likely consumable.

dcaunt:
Yes, you do have to install `carthage` on Travis. Homebrew is the easiest way although it is currently back at version `0.8.0` not the latest `0.9.1`.

eliperkins:
So after a bit of smashing around, I can't seem to get Travis to build the project using Carthage. After a bit of digging around, I found Carthage/Carthage#235, which detailed out how Carthage does not support creating frameworks without a code sign identity. Carthage itself does this [by adding an identity via build script](https://github.com/Carthage/Carthage/blob/12fb7881876b4987584e49d00befcd2a8b6b354d/script/cibuild#L29-L51).
If this is something more that's desired, I can look a bit into how we can make it happen, but I feel like it might be a bit overkill. Let me know what you think!

dfed:
@eliperkins want to just add the following to `.travis.yml`? If that builds, then we're good. It's def overkill, but Valet builds fast enough that I don't see an issue with some overkill.

```
    - xcodebuild -project Valet.xcodeproj -scheme "Valet-iOS" -sdk iphonesimulator -configuration Debug -PBXBuildsContinueAfterErrors=0 ACTIVE_ARCH_ONLY=0 build
    - xcodebuild -project Valet.xcodeproj -scheme "Valet-Mac" -sdk iphonesimulator -configuration Debug -PBXBuildsContinueAfterErrors=0 ACTIVE_ARCH_ONLY=0 build
```

eliperkins:
Done!

dfed:
This looks great! Giving @EricMuller22 a chance to review. If all looks good we'll merge this later today / upload our first Carthage release.

EricMuller22:
LGTM. @dfed will run with this (update docs, etc.) once it's in. Thanks!

PR #44, opened 2015-09-18
App extension only API

Visorox:
Turning this on prevents the project from building with API's that are not available to app extensions and silences the compiler from warning against linking to this framework.

### DIFF
--- a/Valet.xcodeproj/project.pbxproj
+++ b/Valet.xcodeproj/project.pbxproj
@@ -642,6 +642,7 @@
 		26E682811BA8B3F900EFF4EA /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				APPLICATION_EXTENSION_API_ONLY = YES;
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				CURRENT_PROJECT_VERSION = 1;
 				DEBUG_INFORMATION_FORMAT = dwarf;
@@ -666,6 +667,7 @@
 		26E682821BA8B3F900EFF4EA /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				APPLICATION_EXTENSION_API_ONLY = YES;
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				COPY_PHASE_STRIP = NO;
 				CURRENT_PROJECT_VERSION = 1;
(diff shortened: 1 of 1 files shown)


dfed:
Can you attach a screen cap of what this build setting looks like in Xcode? Trying to determine if this will apply to both the Valet iOS and Valet-iOS targets, or if it might also apply to ValetTouchIDTest or our other Mac or Test targets.

For the curious: [Apple's documentation on what kind of APIs are unavailable to app extensions](https://developer.apple.com/library/prerelease/ios/documentation/General/Conceptual/ExtensibilityPG/ExtensionOverview.html#//apple_ref/doc/uid/TP40014214-CH2-SW6).

dfed:
Just cloned your repo and took a look. Looks like this change applies only to `Valet-iOS` and `Valet-Mac`. After playing around with this and looking at some more APIs, I'm pretty sure it should also apply to the `Valet iOS` and `Valet Mac` targets as well. cc @dnkoutso and @mtauraso who know more about build flags than I do.

dnkoutso:
Does this flag need to be set in the podspec as well?

dfed:
If we set it in the Xcode project for `Valet iOS` and `Valet Mac` (which the Podspec consumes/distributes), do we need to set it in the pod spec @dnkoutso? If so, then yes.

Here's my understanding (correct me if I'm wrong): setting the flag  in `Valet iOS` and `Valet Mac` means we  don't screw up when developing locally (and CI catches us as well). Setting the flag in the Podspec means consumers of the pods will see the flag as well, which will remove the warning that @Visorox is seeing in Carthage for CocoaPods users. Not setting the flag in the Podspec will not remove the warning, even if the `Valet iOS` and `Valet Mac` targets have the flag.

@Visorox, don't worry about the podspec changes for now. We can take care of that once your PR is in. Or, if you'd like, we could submit a PR with the change against your branch.

Visorox:
For reference here's the warning I'm getting:
```
ld: warning: linking against dylib not safe for use in application extensions: /Users/Jamie/Documents/Coding/skim.it/sia-iOS/Carthage/Build/iOS/Valet.framework/Valet
```

The reason I didn't set this flag on the other two targets was because Im not sure if this applies to static libraries.

@dfed I'm also not much of a podspec expert and haven't been using cocoapods for months so I'll leave that change up to you. If you're happy merging this in and sorting out the podspec yourself that sounds good.

dfed:
If `APPLICATION_EXTENSION_API_ONLY` will prevent us from using code that is `NS_EXTENSION_UNAVAILABLE`, then we should have it in the static library target. This will only help us not screw up in development, since extensions can't actually link against static libraries – I imagine we won't want our static library and framework targets to diverge.

Once that change is in, I'll merge.

Visorox:
Fair point. I'll add it now

dfed:
2.0.3 has been published with the new build setting.

PR #46, opened 2015-09-22
fixing the license

danielribeiro:
@dfed @EricMuller22 Adding the Copyright to the license file


dfed:
While counter intuitive, this is a template and not where our license is supposed to go. This file has been reviewed by legal. Thanks for looking out, but I'm closing this PR

Issue #33, opened 2015-08-23
Xcode 7 beta: Test environment doesn't write kSecAttrAccessGroup or kSecAttrAccessControl values on keychain items


dfed:
rdar://22395553 has been filed.

dfed:
This causes unit test `test_sharedAccessGroupPrefix_returnsValidValue` to fail

dfed:
If Apple sends the radar back as behaves correctly, we'll need to remove the unit test since it will never work in the simulator. I'll need to create a few signed builds/apps to make sure that shared keychain still works under Xcode 7 if we do that.

dfed:
No word from Apple. Just tested on a signed build. `_sharedAccessGroupPrefix` still works on iOS 9. Going to remove the unit test :cry:.

dfed:
#41 is out. Closing this.

PR #49, opened 2015-10-20
Expose a way to override kSecAttrService

Autoc0diq:
Legacy app may have created shared keychain using a different service than the hardcoded one. This allows to specify it.

### DIFF
--- a/Valet/VALValet.h
+++ b/Valet/VALValet.h
@@ -70,10 +70,13 @@ typedef NS_ENUM(NSUInteger, VALMigrationError) {
 /// @see VALAccessibility
 - (nullable instancetype)initWithIdentifier:(nonnull NSString *)identifier accessibility:(VALAccessibility)accessibility NS_DESIGNATED_INITIALIZER;
 
+- (nullable instancetype)initWithIdentifier:(nonnull NSString *)identifier accessibility:(VALAccessibility)accessibility serviceAttribute:(nonnull NSString *)serviceAttribute NS_DESIGNATED_INITIALIZER;
+
 /// Creates a Valet that reads/writes keychain elements that can be shared across applications written by the same development team.
 /// @param sharedAccessGroupIdentifier This must correspond with the value for keychain-access-groups in your Entitlements file.
 /// @see VALAccessibility
 - (nullable instancetype)initWithSharedAccessGroupIdentifier:(nonnull NSString *)sharedAccessGroupIdentifier accessibility:(VALAccessibility)accessibility NS_DESIGNATED_INITIALIZER;
+- (nullable instancetype)initWithSharedAccessGroupIdentifier:(nonnull NSString *)sharedAccessGroupIdentifier accessibility:(VALAccessibility)accessibility serviceAttribute:(nonnull NSString *)serviceAttribute NS_DESIGNATED_INITIALIZER;
 
 - (nullable instancetype)init NS_UNAVAILABLE;
 + (nullable instancetype)new NS_UNAVAILABLE;
--- a/Valet/VALValet.m
+++ b/Valet/VALValet.m
@@ -209,6 +209,48 @@ OSStatus VALAtomicSecItemDelete(CFDictionaryRef query)
     return [[self class] _sharedValetForValet:self];
 }
 
+- (nullable instancetype)initWithIdentifier:(nonnull NSString *)identifier accessibility:(VALAccessibility)accessibility serviceAttribute:(nonnull NSString *)serviceAttribute;
+{
+    VALCheckCondition(identifier.length > 0, nil, @"Valet requires an identifier");
+    VALCheckCondition(accessibility > 0, nil, @"Valet requires a valid accessibility setting");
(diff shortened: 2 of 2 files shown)


dfed:
Apps migrating to Valet should use `migrateObjectsMatchingQuery:removeOnCompletion:` to bring existing keychain items into Valet, rather than attempting to make Valet work with the existing identifier/service combo.

In case you're curious here's why Valet sets the service internally:
• It ensures that changing Accessibility and storage type (iCloud vs local, etc) metadata works 100% of the time. Updating metadata on existing keychain entities has a ~0.5% failure rate, and the failures are silent. Valet avoids this issue by relying on copying the data from one service to another (which does not fail), rather than updating in place. In fact, our migration code would fail if Valet if the service were set manually, as updating metadata would result in `errSecDuplicateItem` since Valet is not built to update items in place.
• It ensures that each Valet lives in a sandbox – Valets that do not pass the isEqual: test should not be able to stomp one another's values.
• It keeps external adopters of Valet from needing to grapple with multiple unique identifiers (service vs identifier, etc).
• It gives Valet enough control over its ecosystem to ensure that operations will fail only when you access to the keychain is denied.

Hopefully that elucidates why Valet does what it does :). Do let us know if you run into issues using `migrateObjectsMatchingQuery:removeOnCompletion:`.

Autoc0diq:
Thanks for the info. Let me explain why we need this.
We have a legacy app that exposes some data via a shared access group. Unfortunately this app is now maintained and cannot be updated to use Valet in the field. This app was using a certain kAttrService. Without a way to specify the kAttrService when creating a VALValet, there would be no way to even just read that data.
The solution to migrate the data is possible but not ideal. We would have to migrate the data every time we try to access that shared access group.
I think exposing that method in a class extension would be a good compromise.

Autoc0diq:
Side note. None of the tests exercise the initializer with the sharedAccessGroup. Some tests are actually not doing what the test method name is saying i.e.: test_stringForKey_differentAccessGroupFailsToRetrieveString

It also appears that the method _sharedAccessGroupPrefix is returning an empty bundleSeedID when running the tests after trying to add a test for sharedAccessGroup.

dfed:
You can migrate the data in using `migrateObjectsMatchingQuery:removeOnCompletion:`, no? e.g. `[valet migrationObjectsMatchingQuery:{(__bridge id)kAttrService : @"oldServiceName"} removeOnCompletion:YES]`. Valet will then bring that old keychain value into your new Valet, and then there's no need to modify Valet.

The reason none of the tests test the `sharedAccessGroup` initializer is that Xcode's test environment doesn't really work with shared access groups. `_sharedAccessGroupPrefix` returning `nil` in the Xcode 7 test environment is a known issue. See #33.

As for `test_stringForKey_differentAccessGroupFailsToRetrieveString`... good find! I'll look into that. Any others you found?

Autoc0diq:
As I explained, our legacy app still needs the old data in the current place with given kAttrService as such  we cannot migrate the data. We could and leave the data in place by passing NO to removeOnCompletion but we would have to run this migration everytime we run our new app in case the shared data was modified by our legacy app. Seems like a hack when a simple new initializer could be exposed.

dfed:
Ah I didn't realize that you needed to keep the old and new keychain data in sync rather than doing a direct forward migration.

The new initializer violates a central tenant to the Valet architecture (see the explanation above, though I'm happy to elaborate on any of the points), and would give adoptees of Valet rope to hang themselves with. For example, if you were to have a Valet that read the old keychain data and then you attempted to migrate that Valet to a new Valet with the same service but different metadata, your migration would fail.

Migrating the old data each app launch is your best bet.

Autoc0diq:
What type of metadata? Accessibility? I think it would be fine to fail the migration in this case. Is that the only use case?

Wouldn't the same situation already occur if you were to use the existing method migrateObjectsMatchingQuery:removeOnCompletion passing a kAttrService set to the new valet internal service name and different accessibility values?

dfed:
The "metadata" refers to accessibility, Valet type (secure enclave vs iCloud storage vs regular storage), and whether the keychain items are stored with a shared access group or not. Again, the fact that the service name is derived internally is a key component in how Valet is able to guarantee that Valet will only fail to retrieve data when the keychain is inaccessible. It is essential to our sandboxing and threading model, in addition to our migration model.

Moreover, setting the service name manually renders the `identifier` ivar meaningless for Valets that are not created with a shared access group.

`migrateObjectsMatchingQuery:removeOnCompletion:` isn't meant to be used when migration between Valet objects (at least, not externally), that's why we wrote `migrationObjectsFromValet:removeOnCompletion:`. If you're passing in a `kAttrService` that'll match the new valet internal service name (which is prefixed with `VAL_` so as not to conflict with service names that are used by developers in the wild), it's safe to say you're doing it wrong.

In short, the use case you have – interoperability with an app that can not be updated to use Valet – is not a supported case for Valet, nor is it a desired case for Valet. If you'd like Valet to support this use case in your app, I recommend subclassing VALValet and using your subclass, or using a Category on VALValet to let you set the service name and update the baseQuery.

PR #50, opened 2015-10-20
Fix incorrect tests

Autoc0diq:


ValetTests/ValetTests.m
```diff
@@ -174,20 +174,20 @@ - (void)test_stringForKey_differentIdentifierFailsToRetrieveString;
...
     XCTAssertTrue([self.valet setString:self.string forKey:self.key]);
     VALValet *otherValet = [[VALValet alloc] initWithIdentifier:self.valet.identifier accessibility:VALAccessibilityAfterFirstUnlockThisDeviceOnly];
     [self.additionalValets addObject:otherValet];
     
     NSString *string = [otherValet stringForKey:self.key];
-    XCTAssertNil(string, @"Expected string with Key with different access group to be nil but instead it was %@", string);
+    XCTAssertNil(string, @"Expected string with Key with different accessibility to be nil but instead it was %@", string);
 }
 
 - (void)test_stringForKey_differentValetTypeFailsToRetrieveString;
```
dfed: Thanks for fixing! I'd call this test `test_stringForKey_differentAccessGroupFailsToRetrieveString`.

ValetTests/ValetTests.m
```diff
@@ -174,20 +174,20 @@ - (void)test_stringForKey_differentIdentifierFailsToRetrieveString;
...
 {
     XCTAssertTrue([self.valet setString:self.string forKey:self.key]);
-    VALValet *otherValet = [[VALValet alloc] initWithIdentifier:self.valet.identifier accessibility:VALAccessibilityAfterFirstUnlockThisDeviceOnly];
+    VALValet *otherValet = [[VALValet alloc] initWithSharedAccessGroupIdentifier:self.valet.identifier accessibility:VALAccessibilityAlways];
```
dfed: nit: use `self.valet.accessibility` to ensure that `self.valet.accessibility` and `otherValet.accessibility` stay in sync.

ValetTests/ValetTests.m
Autoc0diq: Yes I'll change that. However I think Valet with shared access group will always fail in the simulator anyway so this test may be bogus.

ValetTests/ValetTests.m
Autoc0diq: Good idea.

ValetTests/ValetTests.m
dfed: Just ran some tests locally (been a bit since I've played around with shared access groups in the sim) and you're right. Let's just remove the test

dfed: Actually, I can just remove the test once I merge this. Thanks again for helping clean up our tests!

PR #51, opened 2015-10-21
Add support for remapping keys during migration

Autoc0diq:
When migrating from an old keychain to a Valet, sometimes we may want to remap the old identifiers to different ones.

### DIFF
--- a/Valet/VALValet.h
+++ b/Valet/VALValet.h
@@ -119,9 +119,13 @@ typedef NS_ENUM(NSUInteger, VALMigrationError) {
 /// @see VALMigrationError
 /// @note The keychain is not modified if a failure occurs.
 - (nullable NSError *)migrateObjectsMatchingQuery:(nonnull NSDictionary *)secItemQuery removeOnCompletion:(BOOL)remove;
+/// @param keyMap An optional <NSString, NSString> dictionary with string replacements for kSecAttrAccount values
+- (nullable NSError *)migrateObjectsMatchingQuery:(nonnull NSDictionary *)secItemQuery keyMap:(nullable NSDictionary *)keyMap removeOnCompletion:(BOOL)remove;
 /// Migrates objects from the passed-in Valet into the receiving Valet instance.
 /// @return An error if the operation failed. Error domain will be <code>VALMigrationErrorDomain</code>, and codes will be of type <code>VALMigrationError</code>
 /// @see VALMigrationError
 - (nullable NSError *)migrateObjectsFromValet:(nonnull VALValet *)valet removeOnCompletion:(BOOL)remove;
+/// @param keyMap An optional <NSString, NSString> dictionary with string replacements for kSecAttrAccount values
+- (nullable NSError *)migrateObjectsFromValet:(nonnull VALValet *)valet keyMap:(nullable NSDictionary *)keyMap removeOnCompletion:(BOOL)remove;
 
 @end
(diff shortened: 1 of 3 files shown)


dfed:
Couldn't this be done in a consuming app with:
```
for (NSString *key in fromValet.allKeys) {
    [toValet setObject:[fromValet objectForKey:key] forKey:(keyMap[key] ?: key)];
}

[fromValet removeAllObjects];
```

Though you'd miss `migrateObjectsMatchingQuery:removeOnCompletion:`'s error checking and safe bail out.

Maybe just do the key migration before migrating to the new Valet?

```
for (NSString *key in fromValet.allKeys) {
    if (keyMap[key] != nil) {
        [fromValet setObject:[fromValet objectForKey:key] forKey:keyMap[key]];
        [fromValet removeObjectForKey:key];
    }
}

[toValet migrateObjectsFromValet:fromValet removeOnCompletion:YES];
```

If you're migrating matching query, you could use a staging Valet with a different identifier to import the old tokens, then run migrate again to your final Valet using one of the above snippets.

I don't think this API change solves a common use case in Valet. I'd opt for simpler API than handling one-off cases (especially if the existing API can be used to solve the problem). Might be best to file an issue asking us if we'd be receptive to an API change prior to submitting a PR that changes API.

EricMuller22:
Just a quick follow up: we love hearing how Valet is getting used in the wild, but it's our job to be opinionated about the API. Valet is not intended to be a Swiss Army knife - we think of it more like a scalpel. New apps have a dead simple keychain integration in minutes and even Square Register's legacy keychain code took us just two hours to fully migrate to Valet. We'd rather err on the side of keeping that simplicity in place.

Autoc0diq:
Following up on the code sample above. The solution to use a fromValet to enumerate the keys off a legacy keychain won't work because the kAttrService may have been hardcoded and there's no way to specify one when creating a VALValet (I had another diff for that but it was rejected). So this only option we have is to use migrate method with matching query.

I did look into extending the existing migrate method, add the new behavior in a category. Unfortunately, the ValetDefines.h is not public and neither are the VALAtomicXXX methods. So I am left to essentially redefine my own version of these as well. 

If you have another alternative, I would love to hear it.

The whole point of the migrate method is for encouraging people to switch to using Valet. However if migration forces you to keep the "mistakes" that were made earlier such as specifying keys that are not correct anymore, this defeats the whole purpose.
 
I appreciate the opinion to keep the interface simple. Maybe you can consider putting it into an extension?

dfed:
I believe this flow would work:

Step 1: `migrateObjectsMatchingQuery:removeOnCompletion:` into `fromValet`.
Step 2: Migrate the keys in `fromValet` using the second snippet above
Step 3: Migrate the keys into `toValet` using `migrateObjectsFromValet:removeOnCompletion:`
And then use `toValet` to interact with Keychain in your code.

That said, feel free to write an extension/category in your app. You shouldn't need access to the underlying keychain API to get done what you want to get done. For example, you could write a `- (nullable NSError *)migrateObjectsMatchingQuery:(nonnull NSDictionary *)secItemQuery keyMap:(nullable NSDictionary *)keyMap removeOnCompletion:(BOOL)remove;` extension that just does the steps above.

As for migrating forcing you to keep mistakes... it's quite easy to change the keys once you've migrated your code into Valet. Square Cash did this when we migrated to Valet. The migration code isn't meant to handle all your migration use cases in one fell swoop. It's meant to make it easy to start using Valet – Valet then helps you easily execute the remaining changes you want to make.

The biggest benefit of moving to Valet is you get to stop thinking about how to use the Keychain (which has an abysmal and failure-prone API). Once your data is in Valet, think about how to accomplish what you want using Valet's API (which will not fail, and is pretty simple).

Autoc0diq:
Thanks. That will work.

Issue #52, opened 2015-10-22
removeObjectForKey:options: returns YES when it fails with -34018


Autoc0diq:
I have been chasing an issue when even after I log out successfully, it logs me back in the next time.
(Disclaimer this is with Xcode debugger attached)

- call removeObjectForKey which returns YES, even though VALAtomicSecItemDelete returned -34018
- immediately call object:forKey to check which as expected returns nil, even though VALAtomicSecItemCopyMatching returned -34018
=> I am assuming that the sequence properly removed the key. However it did not. 

So how am I supposed to call canAccessKeychain() after each call to make sure that they actually succeeded?

Autoc0diq:
It appears to work fine without Xcode debugger attached.

dfed:
That's the error for missing entitlements. So it makes sense that getting onto a signed build would fix it.

But it does seem wrong that `removeObjectForKey:` is returning YES when it failed. Re-opening so I can investigate in the morning.

Autoc0diq:
Actually it's the same build, just running it on device without Xcode attached.

dfed:
Yeah it's only checking `errSecInteractionNotAllowed` and not missing entitlements. I'll need to add that to the checks in the morning

Autoc0diq:
Hard to unit test that one ;-)

dfed:
Weird! In any case, the fix is the same :)

dfed:
This issue has been fixed and is consumable in 2.0.4!

PR #53, opened 2015-10-22
removeObjectForKey:options: returns YES when it fails with -34018

dfed:
Fixes issue #52. Props to @Autoc0diq for discovering and reporting!

cc @EricMuller22

### DIFF
--- a/Other/ValetDefines.h
+++ b/Other/ValetDefines.h
@@ -43,3 +43,6 @@
 
 /// Compile flag for building against iOS 9.
 #define VAL_IOS_9_OR_LATER (TARGET_OS_IPHONE && __IPHONE_9_0)
+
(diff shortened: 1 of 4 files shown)


Other/ValetDefines.h
```diff
@@ -43,3 +43,6 @@
 
 /// Compile flag for building against iOS 9.
 #define VAL_IOS_9_OR_LATER (TARGET_OS_IPHONE && __IPHONE_9_0)
+
+/// Error returned from Security API when the application is not entitled to perform the requested action.
+OSStatus const errSecMissingEntitlement = -34018;
```
dfed: Unfortunately this error is not documented in the API (though a quick search online shows that the mapping is correct). Not sure if this is the best place to dump this definition.

EricMuller22:
Gorm

PR #55, opened 2015-10-26
Speed up debugging by logging when a known Apple keychain bug is encountered

efirestone:
I just spent too long debugging this issue, so hopefully this helps someone else in the future.

### DIFF
--- a/Valet/VALValet.m
+++ b/Valet/VALValet.m
@@ -450,6 +450,10 @@ OSStatus VALAtomicSecItemDelete(CFDictionaryRef query)
             keychainData[(__bridge id)kSecValueData] = value;
             
             status = VALAtomicSecItemAdd((__bridge CFDictionaryRef)keychainData, NULL);
+            
+            if (status == errSecMissingEntitlement) {
+                NSLog(@"A 'Missing Entitlements' error occurred. This is likely due to an Apple Keychain bug. As a workaround try running on a device that is not attached to a debugger.\n\nMore information: https://forums.developer.apple.com/thread/4743\n");
+            }
         }
     }, self.lockForSetAndRemoveOperations);
 #else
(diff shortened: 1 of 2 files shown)


Valet/VALValet.m
```diff
@@ -450,6 +450,10 @@ - (BOOL)setObject:(nonnull NSData *)value forKey:(nonnull NSString *)key options
             keychainData[(__bridge id)kSecValueData] = value;
             
             status = VALAtomicSecItemAdd((__bridge CFDictionaryRef)keychainData, NULL);
+            
+            if (status == errSecMissingEntitlement) {
```
dfed: Maybe `VALCheckCondition(status != errSecMissingEntitlement, , A 'Missing Entitlements' error occurred. This is likely due to an Apple Keychain bug. As a workaround try running on a device that is not attached to a debugger.\n\nMore information: https://forums.developer.apple.com/thread/4743)`?

Curious, were you hitting #4 in https://github.com/square/Valet/blob/master/README.md#debugging? Might also be worth giving a shout out to that part of the README.

Thoughts on also logging this in `objectForKey:options:`?

Valet/VALValet.m
```diff
@@ -450,6 +450,10 @@ - (BOOL)setObject:(nonnull NSData *)value forKey:(nonnull NSString *)key options
...
+                NSLog(@"A 'Missing Entitlements' error occurred. This is likely due to an Apple Keychain bug. As a workaround try running on a device that is not attached to a debugger.\n\nMore information: https://forums.developer.apple.com/thread/4743\n");
```
dfed: curious, why the manual `\n`?

efirestone: It just looked better in the console. Otherwise the last line with the forums link gets lost in the next message.

Valet/VALValet.m
efirestone: I resisted using `CheckCondition` for two reasons: (1) The message it logs is prefixed with "Valet API misuse", which isn't actually true (and pretty misleading) in this case, and (2) I was trying to avoid making functional changes and introducing unnecessary risk.

I don't believe this is the same as #4 in the README, unless the README item is actually misdescribed and was this issue. I'm seeing this in Release as well as Debug, and once it starts happening it happens consistently until I run without the debugger attached.

I don't think adding this to `objectForKey:options:` would do anything since I think reading succeeds (it's just writing that fails). I can check again when I get back to my setup that reproduces this since I'm not positive on that.

dfed: Fair re the check condition. Happy to leave it as an NSLog.

After taking a closer look at the apple link forum, this does seem different from #4 in the README. Mind writing up a #5? Doesn't need to be a blocker, but a fast-follow would be appreciated.

As for `objectForKey:options:`, do let me know!

efirestone: Turns out it does happen when reading as well. I've updated the code and the README.

dfed:
Thanks! Merging

Valet/VALValet.m
dfed: Just occurred to me that this error could occur for legitimate reasons (see #2 in the Debugging section of the readme). If this throws anyone for a loop I'll update the log to point to the Debugging section of the readme

PR #62, opened 2015-12-19
Compare kSecUseAuthenticationUI to nil rather than address of to NULL

davidjb:
The current code, with options `[-Werror,-Wtautological-pointer-compare]` set, will produce the following error, as the condition at line 92 of Valet/VALSecureEnclaveValet.m will always evaluate as true:

```
    error: comparison of address of 'kSecUseAuthenticationUI' not equal to a null pointer is always true
```

This PR changes the check to test the value of this variable rather than the address of the pointer, allowing my iOS 9 project to compile correctly.

### DIFF
--- a/Valet/VALSecureEnclaveValet.m
+++ b/Valet/VALSecureEnclaveValet.m
@@ -89,7 +89,10 @@
     NSDictionary *options = nil;
     
 #if VAL_IOS_9_OR_LATER
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wtautological-compare"
     if (&kSecUseAuthenticationUI != NULL) {
+#pragma GCC diagnostic pop
         options = @{ (__bridge id)kSecUseAuthenticationUI : (__bridge id)kSecUseAuthenticationUIFail };
     } else {
 #pragma GCC diagnostic push


dfed:
This change will cause iOS 6 and 7 devices running this code to crash (as soon as they try to access kSecUseAuthenticationUI, the app will BAD_EXEC). Essentially comparing the pioneer to NULL is a quick-and-dirty way to check OS version.

If you'd like to work around the -Wtautological-pointer-compare issue, try wrapping the offending code in:
```objc
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wtautological-pointer-compare"
  offending code
#pragma GCC diagnostic pop
```

davidjb:
Are devices earlier than iOS 9 running the code? It's within a `#if
VAL_IOS_9_OR_LATER` block.

dfed:
VAL_IOS_9_OR_LATER checks build target at compile time. It doesn't check runtime. I'll admit the define is poorly named.

davidjb:
@dfed, sure, gotcha - the check passes due to compiling *for* iOS 9 and above (as well as everything else).  So in that case, I've updated my PR accordingly with your suggestion; that also resolves the issue.

dfed:
Thanks for the fix! Lmk if you use Carthage or Cocoapods. Will push a version bump today if you need it.

davidjb:
Cheers @dfed.  I'm just using Git submodules at this stage but thanks!

Issue #63, opened 2015-12-30
[VALValet canAccessKeychain] spams the device log


omarkilani:
Hey there,

Calling canAccessKeychain causes securityd and the calling process to spam the device log with messages about adding duplicates.

Every time it's called it prints about 4 messages to the log.

I see that this is some sort of "optimization":

https://github.com/square/Valet/blob/master/Valet/VALValet.m#L251

But IMHO I'm not seeing the real issue in calling containsObjectForKey. It probably spends more time logging the error than it would take to check the keychain.

IMHO2: canAccessKeychain isn't really something you call often since you know after launching/being unsuspended whether you're going to have access for that entire runloop (and can take out a background task assertion if you want to stay awake).

Is there something I'm missing? :)

dfed:
Just ran a quick performance test. Writing `canAccessKeychain` as:
```objc
- (BOOL)canAccessKeychain;
{
    NSString *const canaryKey = @"VAL_KeychainCanaryUsername";
    NSString *const canaryValue = @"VAL_KeychainCanaryPassword";
    
    [self setString:canaryValue forKey:canaryKey];
    NSString *const retrievedCanaryValue = [self stringForKey:canaryKey];
    
    return [canaryValue isEqualToString:retrievedCanaryValue];
}
```

slows down performance of the method by 2x. On the Mac target on my machine average speed goes from ~0.034s to ~0.072s. Not to mention it leaves you open to some potentially nasty threading issues.

In order to avoid logging extraneously, setting a keychain item requires that you first check to see if the key/value pair already exists. If it exists, `SecItemUpdate`, if not, `SecItemAdd`. But we're testing if we can access the keychain, so testing to see if the pair already exists is equivalent to running `objectForKey:`, which we can't yet trust.

So maybe the better way to do this is try to short-circuit the manual `SecItemAdd` if we can.
```objc
- (BOOL)canAccessKeychain;
{
    __block BOOL canAccessKeychain = NO;
    VALExecuteBlockInLock(^{
        NSString *const canaryKey = @"VAL_KeychainCanaryUsername";
        NSString *const canaryValue = @"VAL_KeychainCanaryPassword";
        
        NSString *const retrievedCanaryValue = [self stringForKey:canaryKey];
        if ([canaryValue isEqualToString:retrievedCanaryValue]) {
            canAccessKeychain = YES;
            
        } else {
           // Canary value can't be found. Manually add the value to see if we can pull it back out.
[...]

PR #64, opened 2015-12-30
Prevent canAccessKeychain from logging to the console. Improve performance in likely case.

dfed:
Fixes issue #63 

cc @omarkilani @EricMuller22


EricMuller22:
Catching up after skiing all holiday :ski: - looks gorm to me, thanks @omarkilani

PR #65, opened 2016-01-09
Add a sub-Valet which handles ONLY fingerprints, no PIN fallback.

nicwise:
PR / changes which allow for fingerprint-only Secure Enclave to be used - no PIN fallback.

Some info [here](http://fastchicken.co.nz/2016/01/08/valet-keychain-done-right/) as to why.

I dont love the static in the class, for the baseAccessibility, but it needs to be set/stored before the super init it called, and I'm not sure how else to pass it in.

### DIFF
--- a/Valet.xcodeproj/project.pbxproj
+++ b/Valet.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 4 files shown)


nicwise:
Leave this one - it's also not handling the case where you chagnge the fingerprint list, which returns a specific OSStatus which shoudl really be exposed to the calling app. Will revisit.

PR #69, opened 2016-01-21
Differentiate between user cancel and deleted keychain item in VALSecureEnclaveValet

### LINKED ISSUE #67: Differentiate between user cancel and deleted keychain item in VALSecureEnclaveValet.
I am using `VALSecureEnclaveValet` for storing some sensitive data.

My app needs to behave differently for use-case when user cancels the popup and when user disables PIN and then re-enables it.

Since VALSecureEnclaveValet's `objectForKey:userPrompt:` returns nil for both the cases, how do I differentiate?

dfed:
- Valet to 2.2
- Fixes #67 

CC @EricMuller22 @nijindal @nicwise

### DIFF
--- a/Valet.podspec
+++ b/Valet.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'Valet'
-  s.version  = '2.1.0'
+  s.version  = '2.2.0'
   s.license  = 'Apache License, Version 2.0'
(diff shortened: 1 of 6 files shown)


nijindal:
Looks good to merge.

EricMuller22:
Found a crash here, please hold

EricMuller22:
nil in a dictionary - changing userPrompt to nullable made me look, but we could/would have hit this even without the annotation change.
```
id nilValue = nil;
 [someSecureEnclaveValet stringForKey:key userPrompt:nilValue userCancelled:NULL];
```

Issue #72, opened 2016-02-01
Expose whether item was found on VALSecureEnclaveValet


dfed:
We should expose whether an item was found in the `objectForKey:` methods in `VALSecureEnclaveValet` the same way we expose `userCancelled:`. See discussion after https://github.com/square/Valet/issues/67#issuecomment-174199384

cc @nijindal @EricMuller22

dfed:
@nijindal sorry for the slow progress here. Will get to this soon, but I've been overloading by day-to-day work the last week.

dfed:
This fixed is published with v2.2.1. Closing.

PR #73, opened 2016-02-11
Differentiate between an item not being in the keychain and authentication failures

dfed:
Addresses #72 

cc @nijindal @EricMuller22

### DIFF
--- a/Valet.podspec
+++ b/Valet.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'Valet'
-  s.version  = '2.2.0'
+  s.version  = '2.3.0'
   s.license  = 'Apache License, Version 2.0'
   s.summary  = 'Valet lets you securely store data in the iOS or OS X Keychain without knowing a thing about how the Keychain works. It\'s easy. We promise.'
   s.homepage = 'https://github.com/square/Valet'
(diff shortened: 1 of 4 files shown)


dfed:
I'm not sure we can ever tell with 100% certainty that the data has been deleted. It's a guess. The more I'm thinking about this, I think that `userCancelled` should consider `errSecAuthFailed` to be a cancel. `errSecAuthFailed` occurs when the user cancelled the auth by locking the screen, or when the auth fails because the user failed to enter their password correctly (which doesn't feel like a huge stretch to call that a "cancel").

You had [previously said](https://github.com/square/Valet/issues/67#issuecomment-174199384) that "cancel the first authentication popup [and] Subsequent fetch call doesn't trigger an authentication pop up and gives an `errSecAuthFailed` error", but I've been unable to reproduce that behavior using the [VALSecureEnclaveTest](https://github.com/square/Valet/tree/master/ValetTouchIDTest) app  – calling `stringForKey:` after a user cancel does still prompt for TouchID. I noticed in your attached code you waited one runloop after a cancel to pop the dialog – this is likely way too aggressive for LocalAuthentication+Keychain since the TouchID UI requires XPC communication. My best advice is to slow down when talking to a `SecureEnclaveValet` – it may be best talk to a VALSecureEnclave on user-initiated events.

Given the above, I think I'm just going to add `errSecAuthFailed` to the `userCancelled` and close this PR. There are other (possibly recoverable) times when keychain access is hosed, but those should be caught by `canAccessKeychain`. If a customer didn't cancel TouchID (or cause a `errSecAuthFailed`) and you get `nil` back from Valet, I think it makes sense to ask the customer to authenticate again.

dfed:
Closing this PR in favor of #74.

nijindal:
Sounds good @dfed

PR #74, opened 2016-02-16
Treat authentication failures like a user cancel

dfed:
Addresses #72. See discussion in #73 for reasoning behind this approach.

CC @EricMuller22 @nijindal


EricMuller22:
LGTM - @nijindal does this fit your use case (taking into account @dfed's advice that these Touch ID Valets should probably be closer to the main thread/UI than other Valets)?

nijindal:
Sounds good. GTM

PR #76, opened 2016-04-20
Rewrite TouchID test in Swift

EricMuller22:
@dfed


ValetTouchIDTest/ValetTouchIDTestAppDelegate.swift
```diff
@@ -0,0 +1,18 @@
...
+import UIKit
+
+
+@UIApplicationMain
+
+
+final class ValetTouchIDTestAppDelegate : UIResponder, UIApplicationDelegate
+{
+    var window: UIWindow?
+}
```
dfed: missing trailing newline (do we care about this in Swift? I can be swayed either way here)

ValetTouchIDTest/ValetTouchIDTestViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+
+import UIKit
+
+
+final class ValetTouchIDTestViewController : UIViewController
+{
+    // MARK: Properties
+
+    @IBOutlet var textView : UITextView?
+    let secureEnclaveValet : VALSecureEnclaveValet = VALSecureEnclaveValet.init(identifier: "UserPresence", accessControl: VALAccessControl.UserPresence) as VALSecureEnclaveValet!
```
dfed: nit: no need to specify type on the left hand side.
Also, why on earth do you need to `as VALSecureEnclaveValet!`? That really shouldn't be necessary.

The following should do:
`let secureEnclaveValet = VALSecureEnclaveValet.init(identifier: "UserPresence", accessControl: VALAccessControl.UserPresence)`
 
If it doesn't... we should fix our initializers to work better with Swift.

dfed: ohhhhh I think I see what you're doing. Can't you just:
`let secureEnclaveValet = VALSecureEnclaveValet(identifier: "UserPresence", accessControl: VALAccessControl.UserPresence)`

I think your use of `.init` is throwing the compiler for a loop. Just a guess though

ValetTouchIDTest/ValetTouchIDTestViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+    let username : String = "CustomerPresentProof"
+
+
+    // MARK: Outlets
+
+    @IBAction func setOrUpdateItem(sender: UIResponder)
+    {
+        let stringToSet = "I am here! " + NSUUID.init().UUIDString
```
dfed: `NSUUID().UUIDString` should do just fine.

ValetTouchIDTest/ValetTouchIDTestViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+            resultString = password as String!
+        }
+
+        updateTextView([#function, resultString])
+    }
+
+    @IBAction func removeItem(sender: UIResponder)
+    {
+        let removedItem = secureEnclaveValet.removeObjectForKey(username)
+        updateTextView([#function, (removedItem ? "Success" : "Failure")])
```
dfed: TIL! didn't know about `#function`

ValetTouchIDTest/ValetTouchIDTestViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+    @IBAction func setOrUpdateItem(sender: UIResponder)
+    {
+        let stringToSet = "I am here! " + NSUUID.init().UUIDString
+        let setOrUpdatedItem = secureEnclaveValet.setString(stringToSet, forKey: username)
+        updateTextView([#function, (setOrUpdatedItem ? "Success" : "Failure")])
+    }
+
+    @IBAction func getItem(sender: UIResponder)
+    {
+        var userCancelled: ObjCBool = false
```
dfed: curious, does this need to be an `ObjCBool`? If so, why (and we should aim to fix this)?

ValetTouchIDTest/ValetTouchIDTestViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+    @IBAction func containsItem(sender: UIResponder)
+    {
+        let containsItem = secureEnclaveValet.containsObjectForKey(username)
+        updateTextView([#function, (containsItem ? "YES" : "NO")])
+    }
+
+
+    // MARK: Private
+
+    private func updateTextView(messageComponents: [String])
```
dfed: like this a lot. Though I'd use variadic arguments rather than an array.

dfed:
Thanks for giving this a pass! A few qs, but overall looks great!

ValetTouchIDTest/ValetTouchIDTestViewController.swift
EricMuller22: The major issue is that the initializer is nullable so it's returning a `VALSecureEnclaveValet?` What I'm doing is probably poor form, though, I'll think about it a bit more

ValetTouchIDTest/ValetTouchIDTestViewController.swift
EricMuller22: Won't compile as an inout Swift Bool. This was the number one pain point I saw for sure

emish: According to the Swift Compiler ObjCBool is the representation of Objective-C BOOL in Swift because that underlying representation is different in 64 vs. 32-bit. If you want to represent it as Bool in swift, but pass it correctly to ObjC as an inout, you can do

```
let userCancelled: Bool = true
var userCancelledObjC = ObjCBool(userCancelled)
let password = secureEnclaveValet.stringForKey(username, userPrompt: "Use TouchID to retrieve password", userCancelled:&userCancelledObjC)
```

However, this is totally circumventing the point of passing an inout (pointer) to the function, considering that userCancelled was declared as let, it obviously cannot be modified by the function `secureEnclaveValet.stringForKey:userPrompt()`. However, if this isn't an interesting part of the function contract for this call site, it's not a problem.

I would approach the author of `secureEnclaveValet.stringForKey:userPrompt()` and ask why a BOOL pointer is necessary.

ValetTouchIDTest/ValetTouchIDTestViewController.swift
```diff
@@ -0,0 +1,82 @@
...
+    // MARK: Properties
+
+    @IBOutlet var textView : UITextView?
+    let secureEnclaveValet : VALSecureEnclaveValet
+    let username = "CustomerPresentProof"
+
+
+    // MARK: Initializers
+
+    required init?(coder aDecoder: NSCoder)
```
dfed: on Monday let's discuss if it's possible to make these initializers not fail-able (though this may be the best approach). Short of that, this is the way to go for sure.

Some recommended reading: http://www.cocoawithlove.com/blog/2016/01/25/partial-functions-part-one-avoidance.html

ValetTouchIDTest/ValetTouchIDTestViewController.swift
dfed: Good question! We need to communicate to the caller the difference between a user cancelling Touch ID and the item not being in the keychain. Unfortunately, due to bugs in Apple's code, `containsObjectForKey:` isn't guaranteed to give us the right information (see #67 for a detailed discussion); the only way to confirm whether a secure enclave object exists in the keychain is to attempt to fetch it with Touch ID.

In ObjC land, an inout bool was the best approach. In a Swift world, I think we'll want to turn the result into an enum with associated values. @EricMuller22 and I are going to discuss this on Monday at 3:30pm. LMK if you'd like to be part of that meeting :)

dfed:
LGTM

dan_dances

PR #79, opened 2016-06-15
Migrate most of our unit tests to Swift

EricMuller22:
@dfed @brianpartridge @emish


Valet/Valet iOS-Bridging-Header.h
```diff
@@ -0,0 +1,4 @@
+//
+//  Use this file to import your target's public headers that you would like to expose to Swift.
+//
+
```
dfed: we don't need this file unless we think objc is going to pull in swift stuff, which seems unlikely.

Valet/Valet.swift
```diff
@@ -0,0 +1,12 @@
...
+//  Valet
+//
+//  Created by Eric Muller on 4/22/16.
+//  Copyright © 2016 Square, Inc. All rights reserved.
+//
+
+import Foundation
+
+
+
```
dfed: I take it this file is WIP?

ValetTests/ValetTests.swift
```diff
@@ -0,0 +1,742 @@
...
+//
+
+import Foundation
+import XCTest
+
+
+extension NSError
+{
+    // The NSError.code -> VALMigrationError conversion is gross right now:
+    var valetMigrationError: VALMigrationError { return VALMigrationError(rawValue: UInt(self.code))! }
```
dfed: Kinda gross, since this line could crash if `valetMigrationError` is used incorrectly. But... I'm okay with `!`s in test code.

ValetTests/ValetTests.swift
```diff
@@ -0,0 +1,742 @@
...
+}
+
+
+class TestValet: VALValet {}
+
+
+class ValetTests: XCTestCase
+{
+    static let identifier = "valet_testing"
+    let valet = VALValet(identifier: identifier, accessibility: .WhenUnlocked)!
```
dfed: excited to make `VALValet` have a `nonnull` initializer. We'd need to pass in an `Identifier` object which itself has a failable `init` method that takes a string and returns nil if the string passed in has zero length.

ValetTests/ValetTests.swift
```diff
@@ -0,0 +1,742 @@
...
+    let subclassValet = TestValet(identifier: identifier, accessibility: .WhenUnlocked)!
+    let key = "key"
+    let passcode = "topsecret"
```
dfed: dude, everyone knows the best passcode is "12345" ;)

ValetTests/ValetTests.swift
```diff
@@ -0,0 +1,742 @@
...
+    func test_differingSubclassesWithEquivalentConfiguration_areNotEqual()
+    {
+        XCTAssertFalse(valet == subclassValet)
+        XCTAssertFalse(valet === subclassValet)
+    }
+
+    func test_equivalentSubclassesWithEquivalentConfiguration_areEqual()
+    {
+        let secondSubclassValet = TestValet(identifier: ValetTests.identifier, accessibility: .WhenUnlocked)
+        XCTAssertNotNil(subclassValet)
```
dfed: no need of this line, since the following lines will fail if this would fail

ValetTests/ValetTests.swift
```diff
@@ -0,0 +1,742 @@
...
+        }
+    }
+
+    // MARK: containsObjectForKey
+
+    func test_containsObjectForKey()
+    {
+        XCTAssertFalse(valet.containsObjectForKey(key))
+
+        XCTAssert(valet.setString(passcode, forKey: key))
```
dfed: nit: I really like using the specific asserts. e.g. `XCTAssertTrue(x, y)`, `XCTAssertEqual(x, y)`, etc.

Far from a blocker. But to me it feels better to be more specific. More obvious what's happening.

Valet/Valet.swift
EricMuller22: Hmm, I must have started the shim at some point and accidentally committed the new file

ValetTests/ValetTests.swift
EricMuller22: haha see the comment above that line

ValetTests/ValetTests.swift
```diff
@@ -0,0 +1,742 @@
...
+        valet.setString("2", forKey: key)
+        XCTAssertEqual("2", valet.stringForKey(key))
+    }
+
+    func disabled_test_setStringForKey_failsWithInvalidArguments()
+    {
+        // TODO: see if nilValue tap dance tests are even doable in Swift.
+        var nilVar: String?
+        nilVar = nil
+        XCTAssertFalse(valet.setString(nilVar!, forKey: key))
```
dfed: won't this crash?

ValetTests/ValetTests.swift
```diff
@@ -0,0 +1,742 @@
...
+
+        self.waitForExpectationsWithTimeout(5.0, handler: nil)
+    }
+
+    func test_stringForKey_canReadDataWrittenOnAnotherThread()
+    {
+        let setStringQueue = dispatch_queue_create("Set String Queue", DISPATCH_QUEUE_CONCURRENT)
+        let stringForKeyQueue = dispatch_queue_create("String For Key Queue", DISPATCH_QUEUE_CONCURRENT)
+
+        let expectation = self.expectationWithDescription(#function)
```
dfed: style nit: I prefer to leave off `self.` where it isn't necessary

ValetTests/ValetTests.swift
```diff
@@ -0,0 +1,742 @@
...
+    {
+        let setQueue = dispatch_queue_create("Set String Queue", DISPATCH_QUEUE_CONCURRENT)
+        let removeQueue = dispatch_queue_create("Remove Object Queue", DISPATCH_QUEUE_CONCURRENT)
+
+        for _ in 1...50 {
+            dispatch_async(setQueue, { XCTAssert(self.valet.setString(self.passcode, forKey: self.key)) })
+            dispatch_async(removeQueue, { XCTAssert(self.valet.removeObjectForKey(self.key)) })
+        }
+
+        let setQueueExpectation = self.expectationWithDescription("Set String Queue")
```
dfed: For clarity when this fails, might want to call this "\(#function) Set String Queue".

ValetTests/ValetTests.swift
EricMuller22: It's disabled. I'm hoping Mish or someone knows how to do the nilValue tap dance or knows for certain that it's no longer necessary to worry about in Swift-land

dfed: we def don't have to worry about it. one of the beautiful parts of swift is that we can't hax these kinds of things anymore.

ValetTests/ValetTests.swift
```diff
@@ -0,0 +1,734 @@
...
+        ]
+
+        for (key, value) in keyValuePairs {
+            subclassValet.setString(value, forKey: key)
+        }
+
+        XCTAssertNil(valet .migrateObjectsFromValet(subclassValet, removeOnCompletion: false))
+
+        // Both the migration target and the previous Valet should hold all key/value pairs.
+        XCTAssertEqual(5, subclassValet.allKeys().count)
```
dfed: `5` -> `keyValuePairs.allKeys().count`

ValetTests/ValetTests.swift
```diff
@@ -0,0 +1,734 @@
...
+        if VALSecureEnclaveValet.supportsSecureEnclaveKeychainItems() {
+            XCTAssert(valet.setString(passcode, forKey: key))
+            let equivalentValet = VALSecureEnclaveValet(identifier: valet.identifier, accessControl: .DevicePasscode)!
+            XCTAssertNotEqual(valet, equivalentValet)
+            XCTAssertEqual(passcode, valet.stringForKey(key, userPrompt: ""))
+            XCTAssertNil(equivalentValet.stringForKey(key, userPrompt: ""))
+        }
+    }
+
+    @available (*, deprecated)
```
dfed: TIL!

EricMuller22: Yep, that's how you squelch deprecated warnings

ValetTests/ValetTests.swift
```diff
@@ -0,0 +1,734 @@
+//
+//  ValetTests.swift
+//  Valet
+//
+//  Created by Eric Muller on 4/25/16.
+//  Copyright © 2016 Square, Inc. All rights reserved.
```
dfed: delete "All rights reserved".

PR #80, opened 2016-06-16
3.0 Development Branch

EricMuller22:
The one where emuller flails around making things Schwifty
schwifty

cc @dfed @emish @brianpartridge 
#84


danielribeiro:
```xcodebuild: error: The project named "Valet" does not contain a scheme named "Valet Mac". The "-list" option can be used to find the names of the schemes in the project.```

Maybe forgot to mark as a shared scheme?

File: .travis.yml [hunk 1]
```diff
@@ -1,10 +1,9 @@
 language: objective-c
-osx_image: xcode6.4
+osx_image: xcode7.3
```
dfed: since we're going to be using Swift 3, we'll need to go to xcode8

Gemfile
```diff
@@ -1,3 +1,3 @@
 source 'https://rubygems.org'
 
-gem 'cocoapods', '0.39.0'
+gem 'cocoapods', '1.0.1'
```
dfed: you'll need `1.1.0.rc.2` at a minimum to support Swift 3

dfed:
After a way-too-long hiatus, Eric and I are back and working on this PR. There's a lot to do, but we're starting with getting our Swift Valet tests up and running on Xcode 9 GM Seed using Swift 3.2. Once we're building with Swift 3.2, we'll likely change the tests to Swift 4.

Our high-level plan is still the same as we outlined in #84.

Valet/SynchronizableAccessibility.swift
```diff
@@ -0,0 +1,54 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+
+public enum SynchronizableAccessibility: CustomStringConvertible {
```
EricMuller22: possible to protocolize instead?

EricMuller22: also, possibly better spot (maybe not): internal enums to the two classes where we can only expose a subset like this? (synchronizable, secureenclave)

Valet/SynchronizableValet.swift
```diff
@@ -0,0 +1,259 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+
+@available(iOS 7.0.3, macOS 10.9, *)
```
EricMuller22: 💀

Valet/SynchronizableValet.swift
```diff
@@ -0,0 +1,259 @@
...
+/// Reads and writes keychain elements that are synchronized with iCloud (supported on devices on iOS 7.0.3 and later).
```
EricMuller22: comment also outdated 😂

Valet/SynchronizableValet.swift
```diff
@@ -0,0 +1,259 @@
...
+public final class SynchronizableValet: NSObject, KeychainQueryConvertible {
```
EricMuller22: devil's advocate - could we de-duplicate some of this?

ValetTests/SynchronizableTests.swift
```diff
@@ -23,62 +23,59 @@ import Valet
 import XCTest
 
 
-// The iPhone simulator fakes entitlements, allowing us to test the iCloud Keychain (VALSynchronizableValet) code without writing a signed host app.
-#if (arch(i386) || arch(x86_64)) && os(iOS)
+// Until we can write a signed host app on macOS, we can only test the iCloud Keychain (SynchronizableValet) code on iOS.
+#if os(iOS)
```
EricMuller22: I'm a bit nervous about removal of the simulator `#if` logic here - e.g. I _think_ unit tests can be run on device and these will fail unless the build is signed

Valet/SynchronizableValet.swift
EricMuller22: now that we've done this, subclass feels more valid. or.... base Valet but with a different flavor/accessibility

Tests/ValetTests.swift
```diff
@@ -40,12 +40,66 @@ func testEnvironmentIsSigned() -> Bool {
...
+    // MARK: Permutations
+
+    class func permutations(_ identifier: Identifier, shared: Bool = false) -> [Valet] {
+        return Accessibility.allValues().flatMap { accessibility in
+            let flavor: Flavor = .vanilla(accessibility)
+            return shared ? .sharedAccessGroupValet(with: identifier, flavor: flavor) : .valet(with: identifier, flavor: flavor)
+        }
+    }
+
+    class func iCloudPermutations(_ identifier: Identifier, shared: Bool = false) -> [Valet] {
```
dfed: would use `with identifier` rather than just `_ identifier`. Unnamed vars confuse me

Tests/SynchronizableTests.swift
```diff
@@ -41,6 +41,9 @@ class SynchronizableTests: XCTestCase
         }
         
         valet.removeAllObjects()
+        let identifier = SynchronizableTests.identifier
+        let allPermutations = Valet.iCloudPermutations(identifier) + Valet.iCloudPermutations(identifier, shared: true)
+        _ = allPermutations.map { testValet in testValet.removeAllObjects() }
```
dfed: you're looking for `forEach`

Sources/Valet.swift
```diff
@@ -27,11 +27,22 @@ public final class Valet: NSObject, KeychainQueryConvertible {
     
     // MARK: Flavor
     
-    public enum Flavor: Equatable {
+    public enum Flavor: Equatable, CustomStringConvertible {
```
dfed: supernit: `CustomStringConvertible, Equatable` (alphabetical 😅)

Tests/SynchronizableTests.swift
```diff
@@ -116,32 +116,34 @@ class SynchronizableTests: XCTestCase
...
-        XCTAssert(valet.accessibility == .whenUnlocked)
-        let legacyValet = VALSynchronizableValet(identifier: valet.identifier.description, accessibility: VALLegacyAccessibility.whenUnlocked)!
-        
-        let key = "yo"
-        legacyValet.setString("dawg", forKey: key)
-        
-        XCTAssertEqual(legacyValet.string(forKey: "yo"), valet.string(for: key))
+    func test_backwardsCompatibility_withLegacyValet() {
+        for permutation in Valet.iCloudPermutations(valet.identifier) {
+            let legacyValet = VALSynchronizableValet(identifier: permutation.legacyIdentifier, accessibility: permutation.legacyAccessibility)!
```
dfed: likely need to guard `testEnvironmentIsSigned`

Tests/ValetTests.swift
```diff
@@ -40,12 +40,66 @@ func testEnvironmentIsSigned() -> Bool {
...
+        #else
+            XCTFail()
+        #endif
+        return sharedAccessGroupIdentifier
+    }()
+
+    // MARK: Permutations
+
+    class func permutations(_ identifier: Identifier, shared: Bool = false) -> [Valet] {
+        return Accessibility.allValues().flatMap { accessibility in
```
dfed: I think `map` would do here since you never return a `nil` value

Tests/ValetTests.swift
```diff
@@ -65,6 +119,9 @@ class ValetTests: XCTestCase
         
         valet.removeAllObjects()
         anotherFlavor.removeAllObjects()
+        let identifier = ValetTests.identifier
+        let allPermutations = Valet.permutations(identifier) + Valet.permutations(identifier, shared: true)
+        _ = allPermutations.map { testingValet in testingValet.removeAllObjects() }
```
dfed: `forEach`

Sources/Valet.swift
EricMuller22: omg i suck

Tests/SynchronizableTests.swift
EricMuller22: lol

Tests/SynchronizableTests.swift
EricMuller22: caught this in the next commit :(

Sources/Accessibility.swift
```diff
@@ -69,7 +69,7 @@ public enum Accessibility: Int, CustomStringConvertible, Equatable {
         case .afterFirstUnlock:
             accessibilityAttribute = kSecAttrAccessibleAfterFirstUnlock
         case .afterFirstUnlockThisDeviceOnly:
-            accessibilityAttribute = kSecAttrAccessibleWhenUnlockedThisDeviceOnly
+            accessibilityAttribute = kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly
```
dfed: 😫

Sources/SecureEnclaveValet.swift
```diff
@@ -0,0 +1,285 @@
...
+    }
+    
+    // MARK: Public Methods
+    
+    /// - parameter key: A Key used to retrieve the desired object from the keychain.
+    /// - parameter userPrompt: The prompt displayed to the user in Apple's Face ID, Touch ID, or passcode entry UI.
+    /// - returns: The data currently stored in the keychain for the provided key. Returns `nil` if no object exists in the keychain for the specified key, or if the keychain is inaccessible.
+    @available(swift, obsoleted: 1.0)
+    @objc(objectForKey:userPrompt:userCancelled:)
+    public func notforswift_object(for key: Key, withPrompt userPrompt: String, userCancelled: UnsafeMutablePointer<ObjCBool>?) -> Data? {
```
kierajmumick: Maybe r/notforswift/⛔️?

Sources/SecureEnclaveValet.swift
```diff
@@ -0,0 +1,285 @@
...
+    
+    // MARK: Private Properties
+    
+    private let service: Service
+    private let lock = NSLock()
+    private let keychainQuery: [String : AnyHashable]
+}
+
+
+// MARK: - Objective-C Compatibility
```
kierajmumick: Should all the methods below have the notforswift (read: ⛔️) prefix?

dfed: they should! Now fixed.

We had to go with 🚫 instead of ⛔️ since it looks like Xcode doesn't allow composed-character emojis in code.

For reference, 🚫's unicode representation is `U+1F6AB`, and ⛔️'s unicode representation is `U+26D4 U+FE0F`.

README.md
```diff
@@ -56,83 +46,112 @@ Or manually checkout the submodule with `git submodule add git@github.com:Square
...
 
-* Xcode 6.3 or later. Earlier versions of Xcode require [Valet version 1.2.1](https://github.com/square/Valet/tree/1f587a82bed724e75c63f9254287016b51d31ced).
-* iOS 6 or later.
-* OS X 10.10 or later.
+* Xcode 9.0 or later. Earlier versions of Xcode require [Valet version 2.4.2]https://github.com/square/Valet/releases/tag/2.4.2).
+* iOS 9 or later.
+* OS X 10.11 or later.
+
+### Migrating from Valet 2.*
+TODO: write this!
```
EricMuller22: migration guide

README.md
```diff
@@ -56,83 +46,124 @@ Or manually checkout the submodule with `git submodule add git@github.com:Square
...
+```
+
 ```objc
-VALSynchronizableValet *mySynchronizableValet = [[VALSynchronizableValet alloc] initWithIdentifier:@"Druidia" accessibility:VALAccessibilityWhenUnlocked];
+VALValet *const myCloudValet = [VALValet iCloudValetWithIdentifier:@"Druidia" accessibility:VALAccessibilityWhenUnlocked];
 ```
 
-This instance can be used to store and retrieve data that can be retrieved by this app on other devices logged into the same iCloud account with iCloud Keychain enabled. `mySynchronizableValet` can not read or modify values in `myValet` or `mySharedValet` because `mySynchronizableValet` is of a different class type. If iCloud Keychain is not enabled on this device, secrets can still be read and written, but will not sync to other devices.
+This instance can be used to store and retrieve data that can be retrieved by this app on other devices logged into the same iCloud account with iCloud Keychain enabled. `myCloudValet` can not read or modify values in `myValet` or `mySharedValet` because `myCloudValet` is of a different class type. If iCloud Keychain is not enabled on this device, secrets can still be read and written, but will not sync to other devices.
```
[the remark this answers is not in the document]
brridder: A summary table of the APIs, where things are stored for each type, and access interoperability (if any) might help.

Tests/SynchronizableTests.swift
```diff
@@ -23,12 +23,28 @@ import Foundation
 import XCTest
 
 
+extension Valet {
+
+    // MARK: Accessibility -> CloudAccessibility conversion
+
+    var cloudAccessibility: CloudAccessibility {
```
EricMuller22: thoughts here?

Sources/Valet.swift
```diff
@@ -67,33 +67,31 @@ public final class Valet: NSObject, KeychainQueryConvertible {
...
-            
-        } else {
-            let valet = Valet(identifier: identifier, flavor: flavor)
-            identifierToValetMap.setObject(valet, forKey: key)
-            return valet
-        }
+    /// - parameter accessibility: The desired accessibility for the Valet.
+    /// - returns: A Valet that reads/writes keychain elements with the desired accessibility and identifier.
+    public class func valet(with identifier: Identifier, accessibility: Accessibility) -> Valet {
+        return findOrCreate(identifier, flavor: Flavor.vanilla(accessibility))
```
dfed: nit: `.vanilla(accessibility)` (no need for `Flavor`)

Sources/Valet.swift
```diff
@@ -67,33 +67,31 @@ public final class Valet: NSObject, KeychainQueryConvertible {
...
+    }
+
+    /// - parameter identifier: A non-empty string that uniquely identifies a Valet.
+    /// - parameter accessibility: The desired accessibility for the Valet.
+    /// - returns: A Valet (synchronized with iCloud) that reads/writes keychain elements with the desired accessibility and identifier.
+    public class func iCloudValet(with identifier: Identifier, accessibility: CloudAccessibility) -> Valet {
+        return findOrCreate(identifier, flavor: Flavor.iCloud(accessibility))
```
dfed: nit: `.iCloud(accessibility)` (no need for `Flavor`)

Sources/Valet.swift
```diff
@@ -67,33 +67,31 @@ public final class Valet: NSObject, KeychainQueryConvertible {
...
-        if let existingValet = identifierToValetMap.object(forKey: key) {
-            return existingValet
-            
-        } else {
-            let valet = Valet(sharedAccess: identifier, flavor: flavor)
-            identifierToValetMap.setObject(valet, forKey: key)
-            return valet
-        }
+    public class func sharedAccessGroupValet(with identifier: Identifier, accessibility: Accessibility) -> Valet {
+        return findOrCreate(identifier, flavor: Flavor.vanilla(accessibility), sharedAccessGroup: true)
```
dfed: nit: `.vanilla(accessibility)` (no need for `Flavor`)

Sources/Valet.swift
```diff
@@ -67,33 +67,31 @@ public final class Valet: NSObject, KeychainQueryConvertible {
...
+    }
+
+    /// - parameter identifier: A non-empty string that must correspond with the value for keychain-access-groups in your Entitlements file.
+    /// - parameter accessibility: The desired accessibility for the Valet.
+    /// - returns: A Valet (synchronized with iCloud) that reads/writes keychain elements that can be shared across applications written by the same development team.
+    public class func iCloudSharedAccessGroupValet(with identifier: Identifier, accessibility: CloudAccessibility) -> Valet {
+        return findOrCreate(identifier, flavor: Flavor.iCloud(accessibility), sharedAccessGroup: true)
```
dfed: nit: `.iCloud(accessibility)` (no need for `Flavor`)

Sources/Valet.swift
```diff
@@ -67,33 +67,31 @@ public final class Valet: NSObject, KeychainQueryConvertible {
...
-            identifierToValetMap.setObject(valet, forKey: key)
-            return valet
-        }
+    /// - parameter accessibility: The desired accessibility for the Valet.
+    /// - returns: A Valet that reads/writes keychain elements with the desired accessibility and identifier.
+    public class func valet(with identifier: Identifier, accessibility: Accessibility) -> Valet {
+        return findOrCreate(identifier, flavor: Flavor.vanilla(accessibility))
+    }
+
+    /// - parameter identifier: A non-empty string that uniquely identifies a Valet.
```
dfed: "a Valet" -> "a Valet that synchronizes data with iCloud"? (I know we didn't do that before, so nbd)

Tests/SynchronizableTests.swift
dfed: I don't think we need this. All we're doing here is unwrapping the associated value in the `Flavor` enum. Let's just get rid of `Flavor`

`CloudAccessibility(rawValue: accessibility)` will work once we surface a `var accessibility: Accessibility`

Tests/SynchronizableTests.swift
```diff
@@ -23,12 +23,28 @@ import Foundation
...
+        case .vanilla:
+            assert(false)
+            return .always
+        }
+    }
+}
+
+
 @available (iOS 8.2, OSX 10.11, *)
-class SynchronizableTests: XCTestCase
```
dfed: :clap:

Tests/SynchronizableTests.swift
```diff
@@ -23,12 +23,28 @@ import Foundation
...
+            return cloudAccessibility
+        case .vanilla:
+            assert(false)
+            return .always
+        }
+    }
+}
+
+
 @available (iOS 8.2, OSX 10.11, *)
```
dfed: I don't think we need this `@available` anymore

EricMuller22: What are our target minimums now? A few of these may be out of date

Sources/Internal/Configuration.swift
```diff
@@ -22,44 +22,53 @@ import Foundation
...
-    case secureEnclave(SecureEnclave.Flavor)
-    
+    case valet(Accessibility)
+    case iCloud(CloudAccessibility)
+    case secureEnclave(SecureEnclaveAccessControl)
+    case singlePromptSecureEnclave(SecureEnclaveAccessControl)
+
+    // MARK: Public Properties
+
+    public var prettyDescription: String {
```
dfed: nit: `internal` since this entire class is `internal`.

Sources/Internal/Configuration.swift
```diff
@@ -22,44 +22,53 @@ import Foundation
...
+        case .secureEnclave:
+            return "VALSecureEnclaveValet"
+        case .singlePromptSecureEnclave:
+            return "VALSinglePromptSecureEnclaveValet"
         }
     }
     
     // MARK: Internal Properties
     
-    internal var accessability: Accessibility {
```
dfed: 🤦‍♂️

I can't spell 😬

Sources/Internal/Service.swift
```diff
@@ -106,19 +93,12 @@ internal enum Service: CustomStringConvertible, Equatable {
...
-            case let .singlePrompt(desiredAccessControl):
-                accessControl = desiredAccessControl
-                
-            case let .alwaysPrompt(desiredAccessControl):
-                accessControl = desiredAccessControl
-            }
-            
+
+        case let .secureEnclave(accessControl),
+             let .singlePromptSecureEnclave(accessControl):
```
dfed: niiiiiice didn't know we could double `case let`! Is this new to Swift 4??

Tests/SynchronizableTests.swift
dfed: iOS 9.0, and macOS 10.11

dfed:
@EricMuller22 love these last couple commits! We just gotta update the README now :)

Valet.podspec
```diff
@@ -1,15 +1,14 @@
 Pod::Spec.new do |s|
```
dnkoutso: can actually add `test_spec` DSL now to represent your tests by pods which means they would execute when publishing.

Can be in a follow up.

Sources/Internal/Configuration.swift
```diff
@@ -29,7 +29,7 @@ internal enum Configuration: CustomStringConvertible {
 
     // MARK: Public Properties
```
dfed: supernit: `Internal` 😉

README.md
```diff
@@ -156,14 +156,14 @@ First the good news: you will _not_ have to migrate your keychain data when upgr
 
 Now the bad news: the Swift Valet API has slight differences from the Objective-C Valet API. You may have noticed a few of the differences in the sample code above, but here's a rundown of the changes that may affect you.
 
-1. Initializers have changed in both Swift and Objective-C. [See examples above](#Basic-Initialization).
-2. `VALSynchronizableValet` has been replaced by a `Valet`  (or `VALValet` in Objective-C) with an `.iCloud` flavor. [See examples above](#Sharing-Secrets-Across-Devices-with-iCloud).
+1. Initializers have changed in both Swift and Objective-C - both languages use class methods now, which felt more semantically honest (a lot of the time you're not instantiating a new Valet, you're re-accessing one you've already created). [See example usage above](#Basic-Initialization).
```
dfed: love this.

README.md
```diff
@@ -156,14 +156,14 @@ First the good news: you will _not_ have to migrate your keychain data when upgr
...
+2. `VALSynchronizableValet` (which allowed keychains to be synced to iCloud) has been replaced by a `Valet.iCloudValet()`  (or `[VALValet iCloudValet]` in Objective-C). [See examples above](#Sharing-Secrets-Across-Devices-with-iCloud).
```
dfed: `Valet.iCloudValet(with:accessibility:)` and `+iCloudValetWithIdentifier:accessibility:`?

Maybe? Not sure how to properly signify static w/ Swift like we do in Objc

README.md
```diff
@@ -156,14 +156,14 @@ First the good news: you will _not_ have to migrate your keychain data when upgr
...
 6. `migrateObjects(matching:)` and `migrateObjects(from:)` now both return a nonnull [MigrationResult](Sources/MigrationResult.swift#L24).
 7. `VALAccessControl` has been renamed to `SecureEnclaveAccessControl` (`VALSecureEnclaveAccessControl` in Objective-C). This enum no longer references `TouchID`; instead it refers to unlocking with `biometric` due to the introduction of Face ID.
-8. `Valet`, `SecureEnclaveValet`, and `SinglePromptSecureEnclaveValet` are no longer in the same inheritance tree. All three now inherit directly from `NSObject` and use composition to share code.
+8. `Valet`, `SecureEnclaveValet`, and `SinglePromptSecureEnclaveValet` are no longer in the same inheritance tree. All three now inherit directly from `NSObject` and use composition to share code. If you were relying on the subclassing in place before, 1) that might be a code smell 2) consider declaring a protocol for the shared behavior you were expecting to make your migration to Valet 3 easier.
```
dfed: nit: s/subclassing in place before/subclass hierarchy? "in place before" is throwing me a little.

I love the suggestions though

Valet.podspec
EricMuller22: Up to you @dfed I'd be ok with having it happen here.

README.md
EricMuller22: Sure, felt icky trying to shorthand tbh

README.md
EricMuller22: Haha that's how I got Cash compiling so quickly (the protocol)

Valet.podspec
dfed: Well that'd be slick. I'm 💯 for this

README.md
dfed: 🚀

Valet.podspec
EricMuller22: @dnkoutso mind if I grab you mañana?

README.md
```diff
@@ -157,13 +157,13 @@ First the good news: you will _not_ have to migrate your keychain data when upgr
 Now the bad news: the Swift Valet API has slight differences from the Objective-C Valet API. You may have noticed a few of the differences in the sample code above, but here's a rundown of the changes that may affect you.
 
 1. Initializers have changed in both Swift and Objective-C - both languages use class methods now, which felt more semantically honest (a lot of the time you're not instantiating a new Valet, you're re-accessing one you've already created). [See example usage above](#Basic-Initialization).
-2. `VALSynchronizableValet` (which allowed keychains to be synced to iCloud) has been replaced by a `Valet.iCloudValet()`  (or `[VALValet iCloudValet]` in Objective-C). [See examples above](#Sharing-Secrets-Across-Devices-with-iCloud).
+2. `VALSynchronizableValet` (which allowed keychains to be synced to iCloud) has been replaced by a `Valet.iCloudValet(with:accessibility:)`  (or `+iCloudValetWithIdentifier:Accessibility:` in Objective-C). [See examples above](#Sharing-Secrets-Across-Devices-with-iCloud).
```
dfed: capnit: `Accessibility:` -> `accessibility:`

Valet.podspec
dnkoutso: sure!

Sources/Internal/SecItem.swift
```diff
@@ -0,0 +1,219 @@
...
+            return .success
+            
+        } else {
+            ErrorHandler.assert(status != errSecMissingEntitlement, "A 'Missing Entitlements' error occurred. This is likely due to an Apple Keychain bug. As a workaround try running on a device that is not attached to a debugger.\n\nMore information: https://forums.developer.apple.com/thread/4743")
+            
+            return .error(status)
+        }
+    }
+    
+    internal static func delete(itemsMatching query: [String : AnyHashable]) -> Result<OSStatus> {
```
dfed: Supernatural: should probably be `deleteItems(matching query)`

Sources/SecureEnclave.swift
```diff
@@ -0,0 +1,181 @@
...
+            
+        case .error:
+            return false
+        }
+    }
+    
+    /// - parameter key: A Key used to retrieve the desired object from the keychain.
+    /// - parameter userPrompt: The prompt displayed to the user in Apple's Face ID, Touch ID, or passcode entry UI.
+    /// - parameter options: A base query used to scope the calls in the keychain.
+    /// - returns: The data currently stored in the keychain for the provided key. Returns `nil` if no object exists in the keychain for the specified key, or if the keychain is inaccessible.
```
NickEntin: "Returns `nil`..." -> "Returns `.itemNotFound` if no object exists in the keychain for the specified key, or `.userCancelled` if the user cancels the user-presence prompt."  (Same with `string(forKey:withPrompt:options:)` below)

Sources/SecureEnclaveAccessControl.swift
```diff
@@ -0,0 +1,111 @@
...
+//  limitations under the License.
+//
+
+import Foundation
+
+
+@objc(VALSecureEnclaveAccessControl)
+public enum SecureEnclaveAccessControl: Int, CustomStringConvertible, Equatable {
+    /// Access to keychain elements requires user presence verification via Touch ID, Face ID, or device Passcode. Keychain elements are still accessible by Touch ID even if fingers are added or removed. Touch ID does not have to be available or enrolled.
+    /// @version Available on iOS 8 or later, and macOS 10.11 or later.
```
NickEntin: `- version:`, or just remove the `@version`.  Either way, should be consistent across the cases.

Sources/SecureEnclaveValet.swift
```diff
@@ -0,0 +1,285 @@
...
+    
+    // MARK: Private Class Properties
+    
+    private static let identifierToValetMap = NSMapTable<NSString, SecureEnclaveValet>.strongToWeakObjects()
+    
+    // MARK: Initialization
+    
+    @available(*, deprecated)
+    public override init() {
+        fatalError("Do not use this initializer")
```
NickEntin: This is a very vague error.  Should specify which methods to use to initialize instead.

Sources/SecureEnclaveValet.swift
```diff
@@ -0,0 +1,285 @@
...
+    
+    public let identifier: Identifier
+    @objc
+    public let accessControl: SecureEnclaveAccessControl
+    
+    // MARK: Public Methods
+    
+    /// - returns: `true` if the keychain is accessible for reading and writing, `false` otherwise.
+    /// - note: Determined by writing a value to the keychain and then reading it back out. Will never prompt the user for Face ID, Touch ID, or password.
+    @objc
```
NickEntin: Is this necessary here? (Same with on `accessControl` above)

Sources/SecureEnclaveValet.swift
```diff
@@ -0,0 +1,285 @@
...
+    @discardableResult
+    public func set(object: Data, forKey key: String) -> Bool {
+        return execute(in: lock) {
+            return SecureEnclave.set(object: object, forKey: key, options: keychainQuery)
+        }
+    }
+    
+    /// - parameter key: A Key used to retrieve the desired object from the keychain.
+    /// - parameter userPrompt: The prompt displayed to the user in Apple's Face ID, Touch ID, or passcode entry UI.
+    /// - returns: The data currently stored in the keychain for the provided key. Returns `nil` if no object exists in the keychain for the specified key, or if the keychain is inaccessible.
```
NickEntin: "Returns `nil`..." -> "Returns `.itemNotFound`..." (same with `string(forKey:withPrompt:)` below)

Sources/SecureEnclaveValet.swift
NickEntin: Also applies to `Valet` and `SinglePromptSecureEnclaveValet`

README.md
```diff
@@ -56,83 +46,124 @@ Or manually checkout the submodule with `git submodule add git@github.com:Square
...
+This instance can be used to store and retrieve data in the Secure Enclave (available on iOS 8.0 and later and Mac OS 10.11 and later). Each time data is retrieved from this Valet, the user will be prompted to confirm their presence via Face ID, Touch ID, or by entering their device passcode. *If no passcode is set on the device, this instance will be unable to access or store data.* Data is removed from the Secure Enclave when the user removes a passcode from the device. Storing data using SecureEnclaveValet is the most secure way to store data on either iOS or Mac OS.
+
+```swift
+let mySecureEnclaveValet = SinglePromptSecureEnclaveValet.valet(with: Identifier(nonEmpty: "Druidia")!, accessControl: .userPresence)
+```
 
 ```objc
-VALSinglePromptSecureEnclaveValet *mySecureEnclaveValet = [[VALSinglePromptSecureEnclaveValet alloc] initWithIdentifier:@"Druidia" accessControl:VALAccessControlUserPresence];
+VALSinglePromptSecureEnclaveValet *mySecureEnclaveValet = [VALSinglePromptSecureEnclaveValet initWithIdentifier:@"Druidia" accessControl:VALAccessControlUserPresence];
```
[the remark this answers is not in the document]
dfed: good find! That's been there a long time :)

Sources/ErrorHandler.swift
```diff
@@ -0,0 +1,47 @@
...
+import Foundation
+
+
+public final class ErrorHandler {
+    
+    // MARK: Public Static Properties
+    
+    public static var customAssertBody: ((_ condition: Bool, _ message: String, _ file: StaticString, _ line: UInt) -> Void)? = nil
+    
+    public static let defaultAssertBody: (_ condition: Bool, _ message: String, _ file: StaticString, _ line: UInt) -> Void = { condition, message, file, line in
```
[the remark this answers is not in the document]
dfed: fair point. The `defaultAssertBody` does not need to be public.

Sources/Internal/Keychain.swift
```diff
@@ -0,0 +1,401 @@
...
+                
+            case errSecParam:
+                return .invalidQuery
+                
+            default:
+                return .couldNotReadKeychain
+            }
+        }
+        
+        // Now that we have the persistent refs with attributes, get the data associated with each keychain entry.
```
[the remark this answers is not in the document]
dfed: @EricMuller22 do you remember why we pulled out persistent references + attributes first, then queried data?

I want to say one of our previously-supported OSs needed to split this out... we may be able to significantly condense this logic, but I'm honestly afraid of breaking something we don't have a test for here (I tried testing this on iOS 11 and macOS 13 and the data pull passed... not sure if it'd have failed before). Tempted to take the "if it ain't broke" approach here.

Sources/Internal/SecItem.swift
```diff
@@ -0,0 +1,219 @@
...
+    // MARK: Internal Enum
+    
+    internal enum DataResult<SuccessType, FailureType> {
+        case success(SuccessType)
+        case error(FailureType)
+    }
+    
+    internal enum Result<FailureType> {
+        case success
+        case error(FailureType)
```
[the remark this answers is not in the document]
dfed: Good call. No need to be more generic than we need to be. Pulling it out, especially as we're dealing with `internal` properties

Sources/SecureEnclaveAccessControl.swift
dfed: nice find :)

Sources/SecureEnclave.swift
dfed: woof! this was in a bunch of spots.

Sources/SecureEnclaveValet.swift
dfed: fair. Will also add `@available(swift, obsoleted: 1.0)` to these methods to prevent them from being called from Swift.

Sources/SecureEnclaveValet.swift
dfed: Yup! See https://github.com/apple/swift-evolution/blob/master/proposals/0160-objc-inference.md re why we need to explicitly expose to `@objc` now :)

Sources/Internal/Keychain.swift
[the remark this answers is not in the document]
gligorkot: When working with the Keychain, I would also take the "if it ain't broke" approach. Have had too many bad experiences.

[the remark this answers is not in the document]
dfed: Yeah. I'm mostly lamenting the fact that I didn't comment this better when I originally wrote it. I commented the _what_, but not the _why_. Live and learn :/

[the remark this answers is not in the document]
EricMuller22: I don't remember either :(

Tests/MacTests.swift
```diff
@@ -0,0 +1,95 @@
...
+        
+        // We should no longer be able to access the keychain item via the ref.
+        let queryWithVulnerableReferenceAndAttributes = [
+            kSecValueRef as String: vulnerableValueRef,
+            kSecReturnAttributes as String: true
+            ] as CFDictionary
+        XCTAssertEqual(SecItemCopyMatching(queryWithVulnerableReferenceAndAttributes, nil), errSecItemNotFound)
+        
+        // If you add a breakpoint here then manually inspect the keychain via Keychain.app (search for "MacOSVulnTest"), "xctest" should be the only member of the Access Control list.
+        // This is not be the case upon setting a breakpoint and inspecting before the valet.setString(, forKey:) call above.
```
gfontenot: Should this comment have an assertion to go along with it?

Valet.podspec
gfontenot: Did this still need to happen?

ValetTouchIDTest/ValetTouchIDTestViewController.swift
```diff
@@ -18,74 +18,69 @@
...
+        singlePromptSecureEnclaveValet = SinglePromptSecureEnclaveValet.valet(with: Identifier(nonEmpty: "UserPresence")!, accessControl: .userPresence)
+        
         super.init(coder: aDecoder)
     }
-
-
+    
     // MARK: Outlets
-
+
```
gfontenot: Do you mean to be adding this trailing whitespace?

README.md
```diff
@@ -56,83 +46,124 @@ Or manually checkout the submodule with `git submodule add git@github.com:Square
...
+VALValet *const myValet = [VALValet valetWithIdentifier:@"Druidia" accessibility:VALAccessibilityWhenUnlocked];
 ```
 
-To begin storing data securely using Valet, you need to create a VALValet instance with:
+To begin storing data securely using Valet, you need to create a Valet instance with:
 
-* An identifier – a string that is used to identify this Valet.
-* An accessibility value – an enum ([VALAccessibility](Valet/VALValet.h)) that defines when you will be able to store and retrieve data.
+* An identifier – a non-empty string that is used to identify this Valet. The Swift API uses an `Identifier` wrapper class to enforce the non-empty constraint.
+* An accessibility value – an enum ([Accessibility](Sources/Accessibility.swift#L25)) that defines when you will be able to persist and retrieve data.
```
gfontenot: Thoughts on flipping the order between the text and the example? As it is, I see the identifier without knowing what it actually is, and then I need to scroll down to find out.

README.md
```diff
@@ -56,83 +46,124 @@ Or manually checkout the submodule with `git submodule add git@github.com:Square
...
-VALValet *myValet = [[VALValet alloc] initWithIdentifier:@"Druidia" accessibility:VALAccessibilityWhenUnlocked];
+VALValet *const myValet = [VALValet valetWithIdentifier:@"Druidia" accessibility:VALAccessibilityWhenUnlocked];
 ```
 
-To begin storing data securely using Valet, you need to create a VALValet instance with:
+To begin storing data securely using Valet, you need to create a Valet instance with:
 
-* An identifier – a string that is used to identify this Valet.
-* An accessibility value – an enum ([VALAccessibility](Valet/VALValet.h)) that defines when you will be able to store and retrieve data.
+* An identifier – a non-empty string that is used to identify this Valet. The Swift API uses an `Identifier` wrapper class to enforce the non-empty constraint.
```
gfontenot: Maybe add more info here about how the identifier is used? Does it need to be unique? If so, how unique? Maybe offer a suggestion (application name?) for what to use in case they don't know?

README.md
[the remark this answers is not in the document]
gfontenot: I agree. It took me a second to realize you were referring to the instances created above. Also, this seems to imply that `myValet` and `mySharedValet` _can_ read and modify values in eachother. If that's the case then we should make that explicit in the documentation for those flavors.

README.md
```diff
@@ -56,83 +46,124 @@ Or manually checkout the submodule with `git submodule add git@github.com:Square
...
-### Protecting Secrets with Touch ID, Face ID, or device Passcode
+### Protecting Secrets with Face ID, Touch ID, or device Passcode
+
+```swift
+let mySecureEnclaveValet = SecureEnclaveValet.valet(with: Identifier(nonEmpty: "Druidia")!, accessControl: .userPresence)
+```
 
 ```objc
-VALSecureEnclaveValet *mySecureEnclaveValet = [[VALSecureEnclaveValet alloc] initWithIdentifier:@"Druidia" accessControl:VALAccessControlUserPresence];
+VALSecureEnclaveValet *const mySecureEnclaveValet = [VALSecureEnclaveValet initWithIdentifier:@"Druidia" accessControl:VALAccessControlUserPresence];
```
gfontenot: This still needs `alloc`, no? Or should `initWith` be `valetWith`?

ValetTouchIDTest/ValetTouchIDTestViewController.swift
dfed: yeah. We're reverting back to Xcode defaults. Not entirely sure how we ended up missing this whitespace to begin with.

Valet.podspec
dfed: This work is here: https://github.com/square/Valet/pull/120

Tests/MacTests.swift
dfed: There's no good way to test the access control list without doing the above steps, unfortunately. If there were then fixing the ACL vuln would have been a lot simpler 😉

Sources/SecureEnclaveValet.swift
```diff
@@ -0,0 +1,286 @@
...
+    // MARK: Public Class Methods
+    
+    /// - parameter identifier: A non-empty string that uniquely identifies a SecureEnclaveValet.
+    /// - returns: A SecureEnclaveValet that reads/writes keychain elements with the desired flavor.
+    @objc(valetWithIdentifier:accessControl:)
+    public class func 🚫swift_valet(with identifier: String, accessControl: SecureEnclaveAccessControl) -> SecureEnclaveValet? {
+        guard let identifier = Identifier(nonEmpty: identifier) else {
+            return nil
+        }
+        return valet(with: identifier, accessControl: accessControl)
```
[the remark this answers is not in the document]
dfed: I think the guard/return is a bit more clear. Didn't realize that `map` could be used on non-collections to transform a single element (TIL!). Certainly a nifty suggestion though!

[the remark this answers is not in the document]
gfontenot: Ah yeah `Optional.map` is super neat. It's not _exactly_ that you're working with single elements, it's actually available because `Optional` is a 👻 FUNCTOR 👻. The definition is basically:

```swift
extension Optional {
  func map<U>(_ transform: (Wrapped) -> U) -> U? {
    switch self {
      case let .some(value): return transform(value)
      case .none: return .none
    }
  }
}
```

(The [actual implementation](https://github.com/apple/swift/blob/master/stdlib/public/core/Optional.swift#L140-L171) handles throwing functions too).


Note that you _also_ [have access to `Optional.flatMap`](https://github.com/apple/swift/blob/master/stdlib/public/core/Optional.swift#L173-L202), which is equally cool and handy.

Sources/Internal/SecItem.swift
```diff
@@ -0,0 +1,219 @@
...
+internal final class SecItem {
+    
+    // MARK: Internal Enum
+    
+    internal enum DataResult<SuccessType> {
+        case success(SuccessType)
+        case error(OSStatus)
+    }
+    
+    internal enum Result {
```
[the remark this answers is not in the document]
dfed: We originally had just `Result<SuccessType>`, but separated that out into `Result` and `DataResult<SuccessType` in 729b61811442413249fcc3062508994fad1d98b2. We were optimizing for simpler switch/case statements over fewer types. Since this API is entirely internal, having more types isn't a detriment to the overall legibility I don't think.

Sources/SecureEnclave.swift
```diff
@@ -0,0 +1,181 @@
...
+    /// - parameter object: A Data value to be inserted into the keychain.
+    /// - parameter key: A Key that can be used to retrieve the `object` from the keychain.
+    /// - parameter options: A base query used to scope the calls in the keychain.
+    /// - returns: `false` if the keychain is not accessible.
+    @discardableResult
+    internal static func set(object: Data, forKey key: String, options: [String : AnyHashable]) -> Bool {
+        // Remove the key before trying to set it. This will prevent us from calling SecItemUpdate on an item stored on the Secure Enclave, which would cause iOS to prompt the user for authentication.
+        _ = Keychain.removeObject(forKey: key, options: options)
+        
+        switch Keychain.set(object: object, forKey: key, options: options) {
```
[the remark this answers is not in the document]
dfed: I _always_ explicitly iterate over every case using a switch rather than using `==`. I like the compile time safety switching over every case provides; if at some point in the future we add a case, I'll need to determine the right thing to do here before the code compiles again. If I just use `==`, then I can modify this code flow without revisiting it.

Sources/SecureEnclave.swift
```diff
@@ -0,0 +1,181 @@
...
+    }
+    
+    /// - parameter object: A Data value to be inserted into the keychain.
+    /// - parameter key: A Key that can be used to retrieve the `object` from the keychain.
+    /// - parameter options: A base query used to scope the calls in the keychain.
+    /// - returns: `false` if the keychain is not accessible.
+    @discardableResult
+    internal static func set(object: Data, forKey key: String, options: [String : AnyHashable]) -> Bool {
+        // Remove the key before trying to set it. This will prevent us from calling SecItemUpdate on an item stored on the Secure Enclave, which would cause iOS to prompt the user for authentication.
+        _ = Keychain.removeObject(forKey: key, options: options)
```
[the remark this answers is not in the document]
dfed: Since it's internal I do like that we're forcing ourselves to be explicit. We've made our external `removeObject` methods `@discardableResult` at least.

Sources/SecureEnclave.swift
```diff
@@ -0,0 +1,181 @@
...
+        /// User dismissed the user-presence prompt.
+        case userCancelled
+        /// No data was found for the requested key.
+        case itemNotFound
+        
+        // MARK: Equatable
+        
+        public static func ==(lhs: Result<Type>, rhs: Result<Type>) -> Bool {
+            switch lhs {
+            case let .success(lhsResult):
```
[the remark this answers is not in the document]
dfed: Great suggestion!

Sources/SecureEnclaveValet.swift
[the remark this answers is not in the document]
dfed: That's freaking sweet. Thanks for the documentation!!

Is this a standard you've seen used elsewhere? Seems like a really nifty (if slightly unexpected) way to enhance control flow. I half way to steer clear of it because it's a bit unexpected... but the other half of me wants to use the new hotness

Sources/Internal/Service.swift
```diff
@@ -0,0 +1,107 @@
...
+            service = "VAL_\(configuration.description)_initWithIdentifier:accessibility:_\(identifier)_\(configuration.accessibility.description)"
+        case let .sharedAccessGroup(identifier, configuration):
+            service = "VAL_\(configuration.description)_initWithSharedAccessGroupIdentifier:accessibility:_\(identifier)_\(configuration.accessibility.description)"
+        }
+        
+        let configuration: Configuration
+        switch self {
+        case let .standard(_, desiredConfiguration):
+            configuration = desiredConfiguration
+        case let .sharedAccessGroup(_, desiredConfiguration):
```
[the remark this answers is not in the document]
dfed: Good catch!

Sources/SecureEnclaveValet.swift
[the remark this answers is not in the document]
gfontenot: Definitely. I'm a fan of it in general, but I'm a bit of a functional programming nerd. It's especially useful for deferring your need to care if something exists or not until the last possible moment. For example, you could imagine a (n incredibly convoluted) situation where you have an optional string and want to perform some (incredibly convoluted) transformations on it before assigning it to a non-optional property:

```swift
let optString: String? = // maybe I exist, maybe I don't

let transformedString = optString
  .map(uppercase) // uppercase: (String) -> String
  .map(lowercase) // lowercase: (String) -> String
  .map(titlecase) // titlecase: (String) -> String
  .flatMap(ensureNonEmpty) // ensureNonEmpty: (String) -> String?

let string: String = transformedString ?? "default"
```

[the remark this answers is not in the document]
gfontenot: SHAMELESS SELF PROMOTION: I did [a talk about functors at NSNorth 2015](https://github.com/gfontenot/talks/tree/master/Functors) if you're interested in more

Sources/Identifier.swift
```diff
@@ -0,0 +1,45 @@
...
+//
+
+import Foundation
+
+
+public struct Identifier: CustomStringConvertible {
+    
+    // MARK: Initialization
+    
+    public init?(nonEmpty string: String?) {
```
AndrewSB: For maximum schwiftiness, it's possible to create a kind-of dependent types that would enforce this `String` being nonempty. See https://github.com/Ben-G/Validated/blob/master/Validated/Validated.swift for an example if y'all are interested 🙃

Sources/SecureEnclaveValet.swift
```diff
@@ -0,0 +1,286 @@
...
+    
+    // MARK: Private Class Properties
+    
+    private static let identifierToValetMap = NSMapTable<NSString, SecureEnclaveValet>.strongToWeakObjects()
+    
+    // MARK: Initialization
+    
+    @available(*, deprecated)
+    @available(swift, obsoleted: 1.0)
+    public override init() {
```
AndrewSB: Thoughts on marking this as unavailable instead of fatal erring?

README.md
dfed: Great catch! This needs to be `valetWith`

README.md
```diff
@@ -12,35 +12,25 @@ Valet lets you securely store data in the iOS or OS X Keychain without knowing a
...
 
 on iOS:
 
 ```
-platform :ios, '6.0'
+platform :ios, '9.0'
 pod 'Valet'
 ```
 
 on OS X:
```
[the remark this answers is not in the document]
dfed: Whoops! That's been out of date for awhile. Great find 😄

Sources/Accessibility.swift
```diff
@@ -0,0 +1,101 @@
...
+//  limitations under the License.
+//
+
+import Foundation
+
+
+@objc(VALAccessibility)
+public enum Accessibility: Int, CustomStringConvertible, Equatable {
+    /// Valet data can only be accessed while the device is unlocked. This attribute is recommended for data that only needs to be accessible while the application is in the foreground. Valet data with this attribute will migrate to a new device when using encrypted backups.
+    case whenUnlocked = 1
```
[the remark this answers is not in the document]
dfed: Today this is mostly for backwards compatibility. Back in Objective-C a value of `0` could easily be confused with "no value" or nil, so all of our Objective-C enums start with a value of `0`.

Sources/SecureEnclaveValet.swift
dfed: `@available(*, unavailable)` would indeed be better than what we have here. Good thought!

Sources/Identifier.swift
dfed: `Validated` is absurdly nifty. Thanks for the link! However, in this case I think we don't need to go that hard-core. The external-facing API is the same for our `Identifier`, so we could always make this more Schwifty down the line without doing a major version bump :)

dfed:
@AndrewSB thanks for taking a look! I'm going to file an issue for us to add concurrency information to the README. At a high level: use Valet on whatever thread you'd like to use it on. Each Valet is thread safe: it's impossible to concurrently read from & write to a Valet instance. That said, it's good practice to take code that touches the disk (as Keychain does) off the main thread.

README.md
dfed: @EricMuller22 thoughts on this? We originally put the example first to showcase how damned simple it was. But I agree we've traded some simplicity for better compile-time safety, which might make us want to swap here.

Sources/SecureEnclaveValet.swift
AndrewSB: Also, you could mark the function `private` as well, so it isn't accessible outside of the class. Compile time safety that nothing outside the class could use the initializer. I've seen the pattern used a lot in RxSwift

dfed: whoa. IIRC it didn't used to be possible to override without your override matching the inheritance. Will give it a shot. If it works, we'll use it!

dfed: and it worked! Nifty af!

dfed:
PSA: We're planning on merging this PR & publishing v3.0 of Valet on March 2nd! **If you're interested in giving us feedback on this PR, now is the time to do it.** We greatly appreciate everyone above who has spent the time to review + improve this (quite massive) release branch.

Sources/Internal/SecItem.swift
```diff
@@ -0,0 +1,219 @@
...
+        case success
+        case error(OSStatus)
+    }
+    
+    // MARK: Internal Class Properties
+
+    /// Programatically grab the required prefix for the shared access group (i.e. Bundle Seed ID). The value for the kSecAttrAccessGroup key in queries for data that is shared between apps must be of the format bundleSeedID.sharedAccessGroup. For more information on the Bundle Seed ID, see https://developer.apple.com/library/ios/qa/qa1713/_index.html
+    internal static var sharedAccessGroupPrefix: String {
```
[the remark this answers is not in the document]
NickEntin: I assume this will always return the same thing?  If so, we could just make it non-computed property so it only has to run the query once.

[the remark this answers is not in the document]
dfed: It _should_ always return the same thing. In the potential error cases below (I.e the keychain throwing some kind of shit fit) it’s possible that we’d get a bad value once but not later though. It feels safer to compute every time to prevent against an intermittent failure hosing the entire Valet instance for the lifecycle of the object. But I’m also being a bit paranoid – this code _should_ always work, barring the keychain failing at a fundamental level.

gfontenot:
🎉 ✨ 🕺

AndrewSB:
🎈🎉🔒

PR #82, opened 2016-06-16
[DNM,DNR] debugging 3.0 with xcodebuild

danielribeiro:


danielribeiro:
Declining as this pr solves the issue: https://github.com/square/Valet/pull/83

PR #85, opened 2016-06-29
SharedAccessIdentifier Valets should work in the simulator

dfed:
Found out the hard way today that shared access identifier Valets don't work in the simulator. This should fix.

cc @EricMuller22 @kongd

### DIFF
--- a/Valet/VALSecureEnclaveValet.m
+++ b/Valet/VALSecureEnclaveValet.m
@@ -225,7 +225,7 @@ NSString *__nonnull VALStringForAccessControl(VALAccessControl accessControl)
         SEL const backwardsCompatibleInitializer = @selector(initWithSharedAccessGroupIdentifier:accessibility:);
 #if TARGET_IPHONE_SIMULATOR
         /*
-         Access groups do not work on the simulator becuase apps built for the simulator aren't signed.
+         Access groups do not work on the simulator because apps built for the simulator aren't signed.
          Using kSecAttrAccessGroup in the simulator will cause SecItem calls to return -25243 (errSecNoAccessForItem).
          Dropping the kSecAttrAccessGroup key/value pair does not cause problems in development, since all apps can see all keychain items on the simulator.
          */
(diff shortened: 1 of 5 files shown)


Valet/VALSecureEnclaveValet.m
```diff
@@ -223,9 +223,20 @@ - (nullable instancetype)initWithSharedAccessGroupIdentifier:(nonnull NSString *
     self = [super initWithSharedAccessGroupIdentifier:sharedAccessGroupIdentifier accessibility:accessibility];
     if (self != nil) {
         SEL const backwardsCompatibleInitializer = @selector(initWithSharedAccessGroupIdentifier:accessibility:);
+#if TARGET_IPHONE_SIMULATOR
```
dfed: `TARGET_IPHONE_SIMULATOR` is deprecated, replaced by `TARGET_OS_SIMULATOR`. But I'm pretty sure that `TARGET_OS_SIMULATOR` was introduced in Xcode 7, and we're still supporting Xcode 6. Since I didn't want to bump our minimum Xcode version over this, I'm using the old value (which still works)

kongd:
gorm

Valet/VALSecureEnclaveValet.m
efirestone: You can use `||` and check both.

Valet/VALSecureEnclaveValet.m
```diff
@@ -223,9 +223,20 @@ - (nullable instancetype)initWithSharedAccessGroupIdentifier:(nonnull NSString *
...
+        /*
+         Access groups do not work on the simulator becuase apps built for the simulator aren't signed.
```
efirestone: Typo: becuase

dfed: Good one :)

Valet/VALSecureEnclaveValet.m
dfed: Fair. Today `TARGET_IPHONE_SIMULATOR` is `#define`d to be `TARGET_OS_SIMULATOR`... but certainly can't hurt to check both.

Valet/VALSecureEnclaveValet.m
efirestone: Same typo twice more below tok

efirestone:
LGTM

EricMuller22:
lgtm

dfed:
After some investigation on simulator, it looks like this PR doesn't accomplish what it intended. There does not seem to be a way to share keychain items between extensions and host apps in the simulator.

Moreover, unsigned apps (e.g. DEBUG) on device also do not work with shared access groups. In order for to share keychain items between extensions and host in an unsigned environment, the extension and host must use a regular (not a shared access group) Valet. Since Valet can't detect app-level build/signage flags, it is up to adoptees of the library to properly handle this case. I'll make a note to better document this in the README.

PR #87, opened 2016-07-19
[2.2.2] Migration should not fail because `canAccessKeychain` has been called on both Valets

dfed:
cc @EricMuller22


EricMuller22:
GROM

PR #94, opened 2016-09-28
Match nullability specifiers of NSObject initializers

erichoracek:
If you have the `CLANG_WARN_NULLABLE_TO_NONNULL_CONVERSION` warning enabled, the init/new methods on `Valet` produce the following warning:

```
Valet.framework/Headers/VALValet.h:78:1: Conflicting nullability specifier on return types, 'nullable' conflicts with existing specifier 'nonnull'
Valet.framework/Headers/VALValet.h:79:1: Conflicting nullability specifier on return types, 'nullable' conflicts with existing specifier 'nonnull'
```

Additionally, since `[NSObject init]` is now officially declared as not returning nil, I’ve removed the unneeded nil checks from the existing initializers.

### DIFF
--- a/Valet/VALValet.m
+++ b/Valet/VALValet.m
@@ -243,13 +243,12 @@ OSStatus VALAtomicSecItemDelete(__nonnull CFDictionaryRef query)
     VALCheckCondition(accessibility > 0, nil, @"Valet requires a valid accessibility setting");
     
     self = [super init];
-    if (self != nil) {
-        _baseQuery = [[self class] mutableBaseQueryWithIdentifier:identifier accessibility:accessibility initializer:_cmd];
-        _identifier = [identifier copy];
-        _sharedAcrossApplications = NO;
-        _accessibility = accessibility;
-        _lockForSetAndRemoveOperations = [NSLock new];
-    }
+
+    _baseQuery = [[self class] mutableBaseQueryWithIdentifier:identifier accessibility:accessibility initializer:_cmd];
+    _identifier = [identifier copy];
+    _sharedAcrossApplications = NO;
+    _accessibility = accessibility;
+    _lockForSetAndRemoveOperations = [NSLock new];
     
     return [[self class] sharedValetForValet:self];
 }
@@ -260,14 +259,13 @@ OSStatus VALAtomicSecItemDelete(__nonnull CFDictionaryRef query)
     VALCheckCondition(accessibility > 0, nil, @"Valet requires a valid accessibility setting");
     
     self = [super init];
-    if (self != nil) {
-        _baseQuery = [[self class] mutableBaseQueryWithSharedAccessGroupIdentifier:sharedAccessGroupIdentifier accessibility:accessibility initializer:_cmd];
-        
-        _identifier = [sharedAccessGroupIdentifier copy];
-        _sharedAcrossApplications = YES;
-        _accessibility = accessibility;
-        _lockForSetAndRemoveOperations = [NSLock new];
-    }
+
+    _baseQuery = [[self class] mutableBaseQueryWithSharedAccessGroupIdentifier:sharedAccessGroupIdentifier accessibility:accessibility initializer:_cmd];
+    
+    _identifier = [sharedAccessGroupIdentifier copy];
+    _sharedAcrossApplications = YES;
+    _accessibility = accessibility;
+    _lockForSetAndRemoveOperations = [NSLock new];
     
     return [[self class] sharedValetForValet:self];
 }
--- a/Valet.podspec
+++ b/Valet.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'Valet'
-  s.version  = '2.2.2'
+  s.version  = '2.2.3'
   s.license  = 'Apache License, Version 2.0'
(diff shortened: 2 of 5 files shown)


dfed:
Hi @erichoracek, thanks for taking a crack at this! I was looking at doing this with my Xcode 8 adoption PR, but no reason we can't fix the warnings separately. Please note that in order for us to accept this pull request we'll need

- you to sign our [Individual Contributor License Agreement (CLA)](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1).
- a green build in CI

Reviewing the code now.

Valet/VALValet.m
```diff
@@ -237,37 +237,35 @@ + (nullable NSString *)_sharedAccessGroupPrefix;
...
+    _identifier = [identifier copy];
+    _sharedAcrossApplications = NO;
+    _accessibility = accessibility;
+    _lockForSetAndRemoveOperations = [NSLock new];
     
     return [[self class] sharedValetForValet:self];
 }
 
-- (nullable instancetype)initWithSharedAccessGroupIdentifier:(nonnull NSString *)sharedAccessGroupIdentifier accessibility:(VALAccessibility)accessibility;
+- (nonnull instancetype)initWithSharedAccessGroupIdentifier:(nonnull NSString *)sharedAccessGroupIdentifier accessibility:(VALAccessibility)accessibility;
```
dfed: this method can return `nil`, so this needs to remain `nullable`. `VALCheckCondition` returns `nil` if the condition is not met. This comment applies to most of the initializers you have changed – meaning most of the `nullable` -> `nonnull` changes will need to be reverted

Valet/VALValet.m
```diff
@@ -237,37 +237,35 @@ + (nullable NSString *)_sharedAccessGroupPrefix;
...
     self = [super init];
-    if (self != nil) {
-        _baseQuery = [[self class] mutableBaseQueryWithSharedAccessGroupIdentifier:sharedAccessGroupIdentifier accessibility:accessibility initializer:_cmd];
-        
-        _identifier = [sharedAccessGroupIdentifier copy];
-        _sharedAcrossApplications = YES;
-        _accessibility = accessibility;
-        _lockForSetAndRemoveOperations = [NSLock new];
-    }
+
```
dfed: 👍 

I'm a big fan of getting rid of `if (self != nil)` checks where they aren't needed.

Valet/VALValet.m
```diff
@@ -237,37 +237,35 @@ + (nullable NSString *)_sharedAccessGroupPrefix;
 
 #pragma mark - Initialization
 
-- (nullable instancetype)initWithIdentifier:(nonnull NSString *)identifier accessibility:(VALAccessibility)accessibility;
+- (nonnull instancetype)initWithIdentifier:(nonnull NSString *)identifier accessibility:(VALAccessibility)accessibility;
```
dfed: this method can return `nil`, so this needs to remain `nullable`. `VALCheckCondition` returns `nil` if the condition is not met.

Valet/VALValet.m
```diff
@@ -175,7 +175,7 @@ + (nullable NSMutableDictionary *)mutableBaseQueryWithIdentifier:(nonnull NSStri
               } mutableCopy];
 }
 
-+ (nullable NSMutableDictionary *)mutableBaseQueryWithSharedAccessGroupIdentifier:(nonnull NSString *)sharedAccessGroupIdentifier accessibility:(VALAccessibility)accessibility initializer:(nonnull SEL)initializer;
++ (nonnull NSMutableDictionary *)mutableBaseQueryWithSharedAccessGroupIdentifier:(nonnull NSString *)sharedAccessGroupIdentifier accessibility:(VALAccessibility)accessibility initializer:(nonnull SEL)initializer;
```
dfed: good find!

dfed:
We'll also need to update the version in [Valet.podspec](https://github.com/square/Valet/blob/master/Valet.podspec). We should go from `2.2.2` to `2.2.3`.

Valet/VALValet.m
erichoracek: I think this brings up a question regarding the philosophy behind the usage of `VALCheckCondition` here. As-is, it feels like the safety of returning nil in the case of an outright API misuse is not enough to justify the complexity of all consumers of the framework having to handle the potential of a Valet initialization being failable (returning nil). A consumer passing an empty string for a `SharedAccessGroupIdentifier` or a `0` value for `VALAccessibility` are both programmer error, not a runtime error that should be caught and handled appropriately. As such, it seems like these should be `@thrown` `NSInternalInconsistencyException`, rather than early nil returns.

dfed: @EricMuller22 and I have been toying with this kind of redux with our 3.0 release. We'd be open to a PR that migrates some of our `VALCheckCondition` calls to `NSAssert`, which would allow these initializers to become `nonnull`. That said, that kind of change is outside the scope of this PR.

If you'd like us to merge a nullability change along these lines before we get to 3.0, please file an Issue and we can discuss there.

erichoracek:
@dfed Should be ready for re-review—let me know if there's anything else to take care of. Thanks for the prompt review!

erichoracek:
@dfed done!

PR #99, opened 2016-10-28
no ssl bundler

danielribeiro:


danielribeiro:
Travis tries to bundle install before we can patch openssl gem

PR #97, opened 2016-10-26
Add Xcode 8 troubleshooting to Debugging steps

dfed:
cc @EricMuller22 @timdonnelly

Resolves #96

Issue #93, opened 2016-09-26
Make Unit Test suite work on Xcode 8


dfed:
Currently unit tests are broken in Xcode 8 due to a change in how Xcode simulates the keychain in the unit test environment. In Xcode 8, both iOS and Mac unit test targets must both be hosted in a signed app that has permissions to access the keychain.

Here's the total to-do list:
- [ ] Host iOS and Mac unit test targets in a host app
- [ ] Sign the host app with a cert that is checked into the repository
- [ ] Amend README to mention the cert and how to use it
- [ ] Figure out how to add that cert to the keychain in Travis CI using the command line
- [x] Change Travis CI Xcode version to Xcode 8

I'll tackle this in multiple PRs

dfed:
I was able to get iOS tests working without a signed host application – just hosting the tests in an unsigned application worked just fine on the simulator.

macOS was a different beast, however. If macOS tests are hosted in an app, the app _must_ be signed. Furthermore, macOS requires a signed host app in order to use either iCloud syncing or the Secure Enclave. Creating a signed host app presents the following challenges:
- App must be signed with a paid developer account in order to interact with iCloud synced keychain items or the Secure Enclave.
- MPP for the test host app must have individual machines added to it.
- MPP for the test host app must be added to the keychain prior to running tests.

It is possible to run keychain tests on macOS without a host app, however. So for now, we compile out `VALSynchronizableValet` and `VALSecureEnclaveValet` tests on macOS. Developers who want to test `VALSynchronizableValet` and `VALSecureEnclaveValet` in the unit-test environment will need to create a macOS test host app and sign it themselves.

dfed:
Fixed by #102

PR #102, opened 2017-01-16
Enable unit tests and TouchIDTest to run successfully on Xcode 8

dfed:
Allows unit tests to run on Xcode 8. iOS unit tests now run in a host app. Mac unit tests now ignore tests that require entitlements (e.g. SecureEnclave tests). Addresses #93.

Note: this PR is built on top of #97. It will need to be merged first.


dfed:
Tests are failing because I upgraded the Touch ID test app to use Swift 3. Going to bump Travis CI to use Xcode 8 instead of 7. No need for a major version bump because the library still works on Xcode 7.

dfed:
@brianpartridge you already GORM'd #103, but mind giving this guy a quick once-over as well?

brianpartridge:
LGTM

Issue #100, opened 2016-10-28
macOS Secure Enclave Support


brianpartridge:
With touch ID now available on macOS 10.12.1 we should confirm that Valet works with it.  [This line](https://github.com/square/Valet/blob/master/Valet/VALSecureEnclaveValet.h#L25) indicates that we have some macOS support already, but we should verify.  (I would but I'm not running 10.12 yet.)

dfed:
We'll need to include bump `VAL_ACCESS_CONTROL_TOUCH_ID_SDK_AVAILABLE` to include MacOS 10.12.1+

dfed:
Go figure, looks like the Touch ID SDK is available in 10.12+, not just 10.12.1+. ¯\\_(ツ)_/¯

dfed:
This was resolved with #103

PR #103, opened 2017-01-17
Allow TouchID to be use on macOS

dfed:
Enables `SecureEnclaveValet`s to use TouchID on macOS!

Built on top of #102. Addresses #100.

### DIFF
--- a/Valet/VALSecureEnclaveValet.m
+++ b/Valet/VALSecureEnclaveValet.m
@@ -25,7 +25,7 @@
 
 
 /// Compiler flag for building against an SDK where VALAccessControlTouchIDAnyFingerprint and VALAccessControlTouchIDCurrentFingerprintSet are available.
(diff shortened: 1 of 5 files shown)


Valet/VALSecureEnclaveValet.m
```diff
@@ -80,6 +80,18 @@ + (BOOL)_macOSElCapitanOrLater;
 #endif
 }
 
++ (BOOL)_macOSSierraOrLater;
+{
+#if TARGET_OS_MAC && __MAC_10_12
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wtautological-compare"
+    return (&kSecAttrTokenIDSecureEnclave != NULL);
```
brianpartridge: Is this even needed, why aren't the `#if` checks enough?

dfed: Because you may have built against the 10.12 SDK but you're running on 10.11. The compile-time check is SDK rather than OS

brianpartridge: Ah, I've been targeting newer versions only for too long.

PR #105, opened 2017-01-24
Introducing VALSinglePromptSecureEnclaveValet

### LINKED ISSUE #104: kSecUseAuthenticationContext Support
We are developing an application that stores an encryption key on a `VALSecureEnclaveValet` protected by `VALAccessControlTouchIDCurrentFingerprintSet`. We want to protect the key with TouchID but we do not want to request the TouchID everytime we need the key because we use it at almost every screen of our app to decrypt the data displayed on screen. We want to request TouchID once and keep access to the key "authenticated" while our session is alive. Keeping the key in memory for a long time is not an option for us.

We have achieved the behavior we want passing a LAContext object at `kSecUseAuthenticationContext` at the `[VALValet stringForKey:options:] ` options dictionary. But this is a protected method and we cannot call it without modifying your library code. Do you plan to add the possibility to pass a LAContext to `[VALSecureEnclaveValet stringForKey]`? Or could you suggest me a better way to implement this functionality?

Thanks

dfed:
Fixes #104.

Paired with @EricMuller22 on this. @juantrias, please review

Outstanding work:
- [x] Write a toggle into `ValetTouchIDTest` that allows for testing `VALSinglePromptSecureEnclaveValet`
- [x] Update the README to include usage instructions for `VALSinglePromptSecureEnclaveValet`
- [ ] Enable `allKeys` and `removeAllObjects` methods on `VALSinglePromptSecureEnclaveValet` (we will likely punt this to an Issue)

@juantrias, it's worth noting we did not find a workaround for the following issue:
1) Authenticate `VALSinglePromptSecureEnclaveValet` with `TouchIDCurrentFingerprintSet`
2) Change fingerprints
-> Values still show up in existing authenticated `VALSinglePromptSecureEnclaveValet` instance.

I think the best way to get around this is to invalidate the context (call `requirePromptOnNextAccess`) when the app enters the background (e.g. fires `UIApplicationDidEnterBackgroundNotification`). We could not determine a way to detect that fingerprints changed... so it's likely best to invalidate the context whenever the fingerprints _could_ change, i.e. when the app is no longer in the foreground. Not ideal, but neither is the Keychain API ;)

### DIFF
--- a/Valet/VALSecureEnclaveValet.h
+++ b/Valet/VALSecureEnclaveValet.h
@@ -44,11 +44,11 @@ typedef NS_ENUM(NSUInteger, VALAccessControl) {
 };
 
 
-/// Reads and writes keychain elements that are stored on the Secure Enclave (available on iOS 8.0 and later and macOS 10.11 and later) using accessibility attribute VALAccessibilityWhenPasscodeSetThisDeviceOnly. Accessing or modifying these keychain elements will require the user to confirm their presence via Touch ID or passcode entry. If no passcode is set on the device, the below methods will fail. Data is removed from the Secure Enclave when the user removes a passcode from the device. Use the userPrompt methods to display custom text to the user in Apple's Touch ID and passcode entry UI.
+/// Reads and writes keychain elements that are stored on the Secure Enclave (available on iOS 8.0 and later and macOS 10.11 and later) using accessibility attribute VALAccessibilityWhenPasscodeSetThisDeviceOnly. Accessing these keychain elements will require the user to confirm their presence via Touch ID or passcode entry. If no passcode is set on the device, the below methods will fail. Data is removed from the Secure Enclave when the user removes a passcode from the device. Use the userPrompt methods to display custom text to the user in Apple's Touch ID and passcode entry UI.
 /// @version Available on iOS 8 or later, and macOS 10.11 or later.
 @interface VALSecureEnclaveValet : VALValet
 
-/// @return YES if Secure Enclave storage is supported on the current iOS version (8.0 and later).
+/// @return YES if Secure Enclave storage is supported on the current iOS or macOS version (iOS 8.0 and macOS 10.11 and later).
 + (BOOL)supportsSecureEnclaveKeychainItems;
 
 /// Creates a Valet that reads/writes Secure Enclave keychain elements and the specified access control.
--- a/Valet/VALSecureEnclaveValet.m
+++ b/Valet/VALSecureEnclaveValet.m
@@ -19,6 +19,7 @@
 //
 
 #import "VALSecureEnclaveValet.h"
+#import "VALSecureEnclaveValet_Protected.h"
 #import "VALValet_Protected.h"
 
 #import "ValetDefines.h"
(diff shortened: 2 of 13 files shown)


Valet/VALSecureEnclaveValet.h
```diff
@@ -44,11 +44,11 @@ typedef NS_ENUM(NSUInteger, VALAccessControl) {
...
+/// Reads and writes keychain elements that are stored on the Secure Enclave (available on iOS 8.0 and later and macOS 10.11 and later) using accessibility attribute VALAccessibilityWhenPasscodeSetThisDeviceOnly. Accessing these keychain elements will require the user to confirm their presence via Touch ID or passcode entry. If no passcode is set on the device, the below methods will fail. Data is removed from the Secure Enclave when the user removes a passcode from the device. Use the userPrompt methods to display custom text to the user in Apple's Touch ID and passcode entry UI.
 /// @version Available on iOS 8 or later, and macOS 10.11 or later.
 @interface VALSecureEnclaveValet : VALValet
 
-/// @return YES if Secure Enclave storage is supported on the current iOS version (8.0 and later).
+/// @return YES if Secure Enclave storage is supported on the current iOS or macOS version (iOS 8.0 and macOS 10.11 and later).
```
dfed: oops. Missed this documentation change in the last PR.

Valet/VALSecureEnclaveValet.m
```diff
@@ -262,27 +263,7 @@ - (BOOL)canAccessKeychain;
 
 - (BOOL)containsObjectForKey:(nonnull NSString *)key;
 {
-    NSDictionary *options = nil;
```
dfed: This code has been moved down verbatim to `-containsObjectForKey:options:` so that the logic can be shared with `VALSinglePromptSecureEnclaveValet`

Valet/VALSecureEnclaveValet.h
```diff
@@ -44,11 +44,11 @@ typedef NS_ENUM(NSUInteger, VALAccessControl) {
 };
 
 
-/// Reads and writes keychain elements that are stored on the Secure Enclave (available on iOS 8.0 and later and macOS 10.11 and later) using accessibility attribute VALAccessibilityWhenPasscodeSetThisDeviceOnly. Accessing or modifying these keychain elements will require the user to confirm their presence via Touch ID or passcode entry. If no passcode is set on the device, the below methods will fail. Data is removed from the Secure Enclave when the user removes a passcode from the device. Use the userPrompt methods to display custom text to the user in Apple's Touch ID and passcode entry UI.
+/// Reads and writes keychain elements that are stored on the Secure Enclave (available on iOS 8.0 and later and macOS 10.11 and later) using accessibility attribute VALAccessibilityWhenPasscodeSetThisDeviceOnly. Accessing these keychain elements will require the user to confirm their presence via Touch ID or passcode entry. If no passcode is set on the device, the below methods will fail. Data is removed from the Secure Enclave when the user removes a passcode from the device. Use the userPrompt methods to display custom text to the user in Apple's Touch ID and passcode entry UI.
```
dfed: The statement that modifying a keychain item required a Touch ID or passcode entry prompt was invalidated many releases ago – we worked around the issue by deleting the value for key and then setting a new value for key (rather than modifying)

EricMuller22:
Looks gorm to me, sorry about the delay

EricMuller22:
⛷

EricMuller22:
@juantrias thanks for raising #104 - how does this look to you?

dfed:
Last call @juantrias: any feedback here? Will merge by tomorrow AM if we don't hear back.

Issue #113, opened 2017-08-09
Valet error when kSecAttAccount is not a String during keychain migration


kgleong:
[kSecAttrAccount](https://developer.apple.com/documentation/security/ksecattraccount?language=objc) should be a String, but a developer can choose to store an `NSData` object there instead.

Exception:
```
2017-08-09 16:17:20.566 PaySDKDemo[9913:612370] -[__NSCFData isEqualToString:]: unrecognized selector sent to instance 0x6080002a0df0
```

Crash occurs here: https://github.com/square/Valet/blob/2.2.3/Valet/VALValet.m#L426 

Code used to reproduce crash:

```objective-c
// Add an NSData object as the kSecAttrAccount entry
NSDictionary *keychainData = @{(id)kSecAttrService: [NSBundle mainBundle].bundleIdentifier, (id)kSecClass: (id)kSecClassGenericPassword, (id)kSecAttrAccount:[@"bar" dataUsingEncoding:NSUTF8StringEncoding], (id)kSecValueData:[@"foo" dataUsingEncoding:NSUTF8StringEncoding]};

// Manually add the entry above to the keychain
__unused OSStatus status = SecItemAdd((__bridge CFDictionaryRef)keychainData, NULL);

VALValet *keychain = // initialize a Valet keychain

// ** Crash **
[keychain migrateObjectsMatchingQuery:[SQLegacyKeychain migrationQueryWithService:[NSBundle mainBundle].bundleIdentifier] removeOnCompletion:YES];
```

@EricMuller22 @dfed

dfed:
Great find! Should be an easy fix – hopefully we can just sanity check the input. I'm a little embarrassed we weren't already doing that

EricMuller22:
Talked offline (and paired on finding this 😂), but @kgleong is going to follow up with a unit test and fix (probably just an `isKindOfClass` check.)

dfed:
Closing out this issue since #114 fixed it :)

PR #114, opened 2017-08-12
Perform class check on kSecAttrAccount keychain value during migration

kgleong:
* #113
* `(migrateObjectsMatchingQuery: removeOnCompletion:)` assumes that data in `kSecAttrAccount` is a NSString.
* This can cause a crash if the value in kSecAttrAccount is an NSData object, for example.


Valet.xcodeproj/project.pbxproj
```diff
@@ -974,7 +974,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 6.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
```
kgleong: Xcode wouldn't build unless this was > 8.0

dfed: This would be a breaking change, requiring a major version bump. Let's revert this part of the change since it'll affect people using us in submodules?

kgleong:
@dfed that makes sense. I've removed the deployment target change

dfed:
Looks like CI is failing

```
/Users/travis/build/square/Valet/ValetTests/ValetTests.m:684: error: -[KeychainTests test_migrateObjectsMatchingQueryRemoveOnCompletion_withExistingAccountNameNSDataKeychainEntry_doesNotRaiseException] : ((error) == nil) failed: "Error Domain=VALMigrationErrorDomain Code=4 "(null)""
```

Possible that CI is running against a different SDK that didn't have this issue?

kgleong:
ah, looks like the mac test failed.  will look into it, thanks

ValetTests/ValetTests.m
```diff
@@ -90,15 +91,20 @@ - (void)setUp;
...
 
 - (void)tearDown;
 {
-    [super tearDown];
-    
     for (VALValet *const additionalValet in self.additionalValets) {
         [additionalValet removeAllObjects];
     }
+
+    for (NSDictionary *keychainEntry in self.manualKeychainEntries) {
```
kgleong: seems much simpler to do this than some of the other workarounds to reset the keychain:

- https://medium.com/@ryuichi/setup-host-app-for-keychain-tests-in-xcode-8-97222611917e
- https://stackoverflow.com/questions/34907873

dfed: Seems like a reasonable solution! Since this cleanup is needed for only one test may be best to keep the cleanup localized to that test. I'm a big fan of putting the cleanup at the top of the test, so if tearDown is aborted the next run of the test still succeeds.

dfed:
Really good find re Mac/iOS not having the same implementation here. Not surprising at all – there are lovely subtle differences between the two implementations... keychain on iOS and Mac have entirely different implementations under then hood (the code is open source). Thank you again for the fix! Last thing before we get an approval is making the test more resilient to a rerun after failing halfway through the prior run. Almost there!

(Keychain is always more complex than it looks. Always. 😅)

ValetTests/ValetTests.m
kgleong: ah, what about a finally block? I didn't know this existed and it does the same thing i wanted to get out of `tearDown()`: something that is ensured to run regardless of whether an exception is thrown and execution stops.

```objective-c
NSDictionary *keychainEntry = <..>

@try {
.. all the code
}
@finally {
  SecItemDelete(keychainEntry);
}
```

kgleong:
where does one find the open source code?  i remember @EricMuller22 entering some magical google search term...

ValetTests/ValetTests.m
dfed: I'm throwing straw men out there now (but I have been bit by these in the past): `finally` won't prevent against
- Hitting cmd+. in the middle of the test
- Another test polluting the keychain

Normally I don't protect against these kinds of issues in my tests. But since keychain is both persistent and a black box I like to control for as many factors as I can in each test.

dfed:
I tend to google "Apple open source secitem" 🙂

Here's some of it: https://opensource.apple.com/source/Security/Security-55471/sec/Security/SecItem.c

ValetTests/ValetTests.m
kgleong: ah gotcha.  thanks for all the tips and the open source search terms

kgleong:
@dfed I placed the `SecItemDelete` back in its previous place (as early as possible in the test)

dfed:
@EricMuller22 want to merge/release? I'll get to it tomorrow if not.

EricMuller22:
LGTM, I'll let you handle merge/release though @dfed, going to be in meetings all day 😞

kgleong:
Posted this SO question https://stackoverflow.com/questions/47602876 to see if anyone knows why OSX and iOS handle NSData objects in `kSecAttrAccount` differently.  fyi @EricMuller22 @dfed

PR #116, opened 2017-09-27
FIX [VALSinglePromptSecureEnclaveValet containsObjectForKey] on TouchID fingerprints changes 

juantrias:
`[VALSinglePromptSecureEnclaveValet containsObjectForKey]` is returning `YES` after the key has been removed from the Keychain due to TouchID fingerprints changes, after adding or removing a fingerprint. The reason is that the `LAContext` instance kept by `VALSinglePromptSecureEnclaveValet` has the key cached. So, for `containsObjectForKey`, we need to query the Keychain without using this instance, as in regular `VALSecureEnclaveValet`.

Additionally, although this does not affect the functionality, we only need to pass the `LAContext` instance when reading keys from the Keychain, not when adding or removing keys

### DIFF
--- a/Valet/VALSinglePromptSecureEnclaveValet.m
+++ b/Valet/VALSinglePromptSecureEnclaveValet.m
@@ -52,29 +52,14 @@
 
 #pragma mark - VALValet
 
-- (BOOL)setObject:(nonnull NSData *)value forKey:(nonnull NSString *)key;
-{
-    return [self setObject:value forKey:key options:[self _contextOptions]];
-}
-
 - (nullable NSData *)objectForKey:(nonnull NSString *)key;
 {
     return [self objectForKey:key options:[self _contextOptions] status:nil];
 }
 
-- (BOOL)setString:(nonnull NSString *)string forKey:(nonnull NSString *)key;
-{
-    return [self setString:string forKey:key options:[self _contextOptions]];
-}
-
 - (nullable NSString *)stringForKey:(nonnull NSString *)key;
 {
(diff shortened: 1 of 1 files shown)


dfed:
> after the key has been removed from the Keychain due to TouchID fingerprints changes

Great find! We'll need to make sure to adopt similar changes in #80.

> Additionally, although this does not affect the functionality, we only need to pass the LAContext instance when reading keys from the Keychain, not when adding or removing keys.

Fascinating! Do you have documentation to support this?

I'm going to run this change though its paces in our TouchIDTest app. If it looks good, I'll merge and release an update :)

dfed:
This looks good! Will merge when CI goes green. Updating #80 now. Thanks again for this fix!!

dfed:
Just updated #80 with 8f83921d3da8431596969e3f9b8258d3201acd15. The next release of Valet won't regress this :)

dfed:
2.4.2 has been published with this change. Thanks again!

juantrias:
Thank you for the quick reply! 

> Additionally, although this does not affect the functionality, we only need to pass the LAContext instance when reading keys from the Keychain, not when adding or removing keys.

I haven't found documentation to support this, just my testing and this transcript from a WWDC Session http://asciiwwdc.com/2014/sessions/711

> Authentication is the new component, and to specify authentication you have to provide policy.
> 
> And policy defines what authentication methods have to be satisfied or done before the Keychain item is decrypted and returned back to your application

Official docs are ambiguous, i.e. for kSecAccessControlTouchIDCurrentSet they say "Constraint to **access** an item with Touch ID for currently enrolled fingers"

PR #120, opened 2018-01-26
Update to use CocoaPods 1.4.0 and test specs

dnkoutso:


Gemfile
```diff
@@ -1,4 +1,3 @@
 source 'https://rubygems.org' do
-  gem 'cocoapods', '1.0.0'
-  gem 'activesupport', '~> 4.0'
```
dfed: 👏

dnkoutso:
@dfed seems probably an issue with minimum version of iOS 6.0...might be a cocoapods issue will investigate.

dfed:
@dnkoutso if you rebase master this may just start working. Minimum version is now iOS 9.0.

dnkoutso:
Done! 🤞

Valet.podspec
```diff
@@ -10,5 +10,13 @@ Pod::Spec.new do |s|
   s.public_header_files = 'Sources/*.h'
   s.frameworks = 'Security'
   s.ios.deployment_target = '9.0'
-  s.osx.deployment_target = '10.11'
+  s.macos.deployment_target = '10.11'
+
+  s.test_spec 'Tests' do |test_spec|
+    test_spec.ios.requires_app_host = true
```
dnkoutso: 1.4.0 can create you a temp app host :)

Valet.podspec
```diff
@@ -10,5 +10,13 @@ Pod::Spec.new do |s|
   s.public_header_files = 'Sources/*.h'
   s.frameworks = 'Security'
   s.ios.deployment_target = '9.0'
-  s.osx.deployment_target = '10.11'
+  s.macos.deployment_target = '10.11'
```
dnkoutso: this is 2018.

Valet.podspec
```diff
@@ -10,5 +10,13 @@ Pod::Spec.new do |s|
...
+
+  s.test_spec 'Tests' do |test_spec|
+    test_spec.ios.requires_app_host = true
+    test_spec.ios.source_files = 'Tests/**/*.{h,m,swift}'
+    test_spec.ios.exclude_files = 'Tests/MacTests.swift'
+    test_spec.macos.source_files = 'Tests/**/*.{h,m,swift}'
+    test_spec.pod_target_xcconfig = { 'SWIFT_OBJC_BRIDGING_HEADER' => '${PODS_TARGET_SRCROOT}/Tests/ValetTests-Bridging-Header.h' }
```
dnkoutso: always use `PODS_TARGET_SRCROOT` as its relative to where the files are placed.

dnkoutso:
@dfed only one test fails on iOS:

```
    - NOTE  | [iOS] xcodebuild:  /Users/travis/build/square/Valet/Tests/SecItemTests.swift:30: error: -[Valet_Unit_Tests.SecItemTests test_sharedAccessGroupPrefix_findsPrefix] : XCTAssertEqual failed: ("FAKETEAMID") is not equal to ("9XUJ7M53NG") - 
```

I also had it locally fail to me I guess its because the temp app host is generated?

dnkoutso:
Yes this fails because the app host does not have `9XUJ7M53NG` as its `DEVELOPMENT_TEAM` set.

dfed:
@dnkoutso we have already generated our own test host apps: on iOS we have an app with bundle ID `com.squareup.Valet-iOS-Test-Host-App` and macOS we have an app with bundle ID `com.squareup.Valet-macOS-Test-Host-App`. Could we re-use the existing test hosts?

dnkoutso:
Unfortunately not during `lint` which will just generate a space as if someone is consuming `Valet`.

I could probably circumvent the test to be honest since you will be getting coverage anyway from your normal `.xcodeproj`

Tests/SecItemTests.swift
```diff
@@ -27,7 +27,9 @@ class SecItemTests: XCTestCase {
 
     func test_sharedAccessGroupPrefix_findsPrefix() {
         #if os(iOS)
-            XCTAssertEqual(SecItem.sharedAccessGroupPrefix, "9XUJ7M53NG")
+            #if !COCOAPODS
```
dnkoutso: @dfed not the best but it will do I think for now. You are already getting good coverage from your xcodeproj. Alternatively I can exclude this file also from the test spec.

dnkoutso: it seems it would be nice if I expand the `app_host` DSL to allow you to specify certain attributes similar to `pod_target_xcconfig` but it will be a while...

dfed: This is fine I think. Let's add a good comment as to why we're only testing this in the non-cocoapods case.

Valet.podspec
```diff
@@ -10,5 +10,16 @@ Pod::Spec.new do |s|
   s.public_header_files = 'Sources/*.h'
   s.frameworks = 'Security'
   s.ios.deployment_target = '9.0'
-  s.osx.deployment_target = '10.11'
+  # s.macos.deployment_target = '10.11'
```
dnkoutso: oops

Tests/SecItemTests.swift
dnkoutso: done

dnkoutso:
@dfed got a green build! This will now run tests when you publish or lint which is always good. If it gives you problems just ping me or just delete the `test_spec` in the DSL and I can get it back on.

One more step eventually would be to remove the `xcodeproj` entirely that you manage and just generate one from the podspec (including all tests) for you to work on on demand. No merge conflicts or manuakl `.xcodeproj` management.

.travis.yml
```diff
@@ -5,4 +5,4 @@ before_script:
 script:
     - xcodebuild -project Valet.xcodeproj -scheme "Valet iOS" -sdk iphonesimulator -destination "platform=iOS Simulator,OS=11.0,name=iPhone 7" -configuration Debug -PBXBuildsContinueAfterErrors=0 build test
     - xcodebuild -project Valet.xcodeproj -scheme "Valet Mac" -sdk macosx10.13 -configuration Debug -destination "platform=OS X" -PBXBuildsContinueAfterErrors=0 build test
-    - pod lib lint --verbose --fail-fast
+    - bundle exec pod lib lint --verbose --fail-fast
```
dnkoutso: also added this.

Gemfile
```diff
@@ -1,3 +1,3 @@
 source 'https://rubygems.org' do
-  gem 'cocoapods', '1.3.1'
+  gem 'cocoapods', '~> 1.4'
```
dfed: ❤️

dfed:
LGTM!

Valet.podspec
```diff
@@ -10,5 +10,16 @@ Pod::Spec.new do |s|
...
+  s.macos.deployment_target = '10.11'
+
+  s.test_spec 'Tests' do |test_spec|
+    test_spec.ios.requires_app_host = true
+    test_spec.ios.source_files = 'Tests/**/*.{h,m,swift}'
+    test_spec.ios.exclude_files = 'Tests/MacTests.swift'
+    test_spec.macos.source_files = 'Tests/**/*.{h,m,swift}'
+    test_spec.pod_target_xcconfig = {
+      'SWIFT_OBJC_BRIDGING_HEADER' => '${PODS_TARGET_SRCROOT}/Tests/ValetTests-Bridging-Header.h',
+      'CLANG_WARN_UNGUARDED_AVAILABILITY' => 'YES'
```
dnkoutso: @dfed in Xcode 9 this is defaulting to `YES_AGGRESSIVE` I think but in your project you only have it to `YES` so you do not get any warnings during compilation of your test sources.

dfed: Interesting. I guess we should do another pass to change our Xcode project to`YES_AGGRESSIVE`?

dnkoutso: yes we can do in a follow up. so we can fix the code and then remove this line from the podspec.

dnkoutso:
green! shall I merge?

dfed:
Please do @dnkoutso!

PR #125, opened 2018-03-03
Explicitly mention the 'use_frameworks!' requirement when pulling in Valet via Cocoapods

dfed:
Now that we're written in Swift, not having `use_frameworks!` in the Podfile would cause our adoptees' builds to fail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ on iOS:
 
 ```
 platform :ios, '9.0'
+use_frameworks!
 pod 'Valet'
 ```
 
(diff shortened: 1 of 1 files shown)


brianpartridge:
Maybe call out that this is required in a sentence too. Not a blocker but would be nice for people on auto pilot adding the pod.

brianpartridge:
LGTM either way

dfed:
Just took a look at a few other Swift repos: no one provides an explanation. I think we're okay as-is.

PR #127, opened 2018-03-04
Add watchOS support

### LINKED ISSUE #40: watchOS Support
If it does not currently, it would be nice to use this library on watchOS 2 (and update the Podspec file to declare this support). I’m happy to help do this, just putting this here to see if anyone else is already working on this.

dfed:
This PR resolves #40.

In order to properly test that our code works on watch, I created a watch test app that scans for all `XCTestCase` `test` methods and runs them. Tests passed on the first go!

The only current Valet feature that doesn't work on watchOS is the `SinglePromptSecureEnclaveValet`, since the `LocalAuthentication` library never made it to watchOS.


dfed:
Just filed #128 to get help running the watch tests in CI. I think we're okay to merge as-is.

Valet.podspec
```diff
@@ -11,19 +11,8 @@ Pod::Spec.new do |s|
   s.frameworks = 'Security'
   s.ios.deployment_target = '9.0'
   s.tvos.deployment_target = '9.0'
+  s.watchos.deployment_target = '2.0'
   s.macos.deployment_target = '10.11'
 
-  s.test_spec 'Tests' do |test_spec|
```
dfed: @dnkoutso do you know how to get the test spec to ignore `watchos`? Since the watch doesn't have the XCTest framework (or really any test infra), we need to skip it. I tried including `test_spec.watchos.exclude_files = 'Tests/**/*.{h,m,swift}'` below [in a previous commit](https://github.com/square/Valet/pull/127/commits/ed0355a2a9efad28fd3fc7041c43a3174b707f48), but I still end up with the following error: `/Users/travis/Library/Developer/Xcode/DerivedData/App-azgcswumbhaldvdurgqjunerahug/Build/Products/Debug-watchsimulator/Valet-Unit-Tests.xctest: bundle format unrecognized, invalid, or unsuitable`

You can see the full error here: https://api.travis-ci.org/v3/job/348993626/log.txt

dfed: @dnkoutso could I get around this by adding a dummy test file for the watchOS target?

dfed: Nope. And just saw https://github.com/CocoaPods/CocoaPods/issues/7009

Valet watchOS Test Host App Extension/ExtensionDelegate.swift
```diff
@@ -0,0 +1,65 @@
...
+
+class ExtensionDelegate: NSObject, WKExtensionDelegate {
+
+    func applicationDidFinishLaunching() {
+        // Perform any final initialization of your application.
+
+        XCTestCase.runAllTests()
+    }
+
+    func applicationDidBecomeActive() {
```
martinmroz: Maybe kill this boilerplate too?

Tests/SecItemTests.swift
```diff
@@ -20,13 +20,15 @@
...
+#if !os(watchOS)
+    import XCTest
+#endif
 
 
 class SecItemTests: XCTestCase {
 
     func test_sharedAccessGroupPrefix_findsPrefix() {
-        #if os(iOS)
+        #if !os(macOS)
```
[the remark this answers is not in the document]
dfed: Good idea! I think we have accidental land mines elsewhere by not having `else` statements. But it can’t hurt to be explicit af

Tests/XCTest+watchOS.swift
```diff
@@ -0,0 +1,262 @@
...
+
+#if os(watchOS)
+    import Foundation
+    import ObjectiveC
+
+
+    // MARK: - XCTestCase
+
+    /// A bare-bones re-implementation of XCTestCase for running tests from the watch extension.
+    @objcMembers
```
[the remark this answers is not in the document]
dfed: I was really happy to discover that this flag applies to subclasses too! Means that we don’t have to add `@objc` to every test to get it to run

Tests/XCTest+watchOS.swift
```diff
@@ -0,0 +1,262 @@
...
+    class XCTestCase: NSObject {
+
+        // MARK: Class Internal
+
+        class func runAllTests() {
```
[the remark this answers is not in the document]
dfed: TIL!

Tests/XCTest+watchOS.swift
```diff
@@ -0,0 +1,262 @@
...
+            assertionFailure(message, file: file, line: line)
+        }
+    }
+
+    func XCTFail(_ message: String = "", file: StaticString = #file, line: UInt = #line) {
+        assertionFailure(message, file: file, line: line)
+    }
+
+    func measure(file: StaticString = #file, line: Int = #line, block: () -> Void) {
+        // We don't need to do anything here.
```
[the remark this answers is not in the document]
dfed: Fair point. Only reason we don’t need anything here is we aren’t using it yet. I’ll add a fatal error with a todo instead of a cryptic comment

Valet.podspec
```diff
@@ -1,8 +1,8 @@
 Pod::Spec.new do |s|
   s.name     = 'Valet'
-  s.version  = '3.1.3'
+  s.version  = '3.2.0'
   s.license  = 'Apache License, Version 2.0'
-  s.summary  = 'Securely store data in the iOS, tvOS, or macOS Keychain without knowing a thing about how the Keychain works. It\'s easy. We promise.'
+  s.summary  = 'Securely store data on iOS, tvOS, watchOS, or macOS without knowing a thing about how the Keychain works. It\'s easy. We promise.'
```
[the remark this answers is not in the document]
dfed: Can you tell there’s a character limit? 😅

Tests/SynchronizableTests.swift
```diff
@@ -72,8 +74,8 @@ class CloudTests: XCTestCase
             return
         }
         let otherValet = Valet.iCloudValet(with: valet.identifier, accessibility: accessibility)
-        XCTAssert(valet == otherValet)
-        XCTAssert(valet === otherValet)
+        XCTAssertTrue(valet == otherValet)
```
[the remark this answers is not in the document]
dfed: Actually, looking back at this, I think the reason I went with `AssertTrue` is having
```
XCTAssertTrue(valet == otherValet)
XCTAssertTrue(valet === otherValet)
```
makes it clearer why I have two lines there. The following feels a little harder to parse:
```
XCTAssertEqual(valet, otherValet)
XCTAssertTrue(valet === otherValet)
```

Maybe I'm crazy?

gfontenot:
Anything I can do to help push this through?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "Tests/xctest-watchOS"]
+	path = Tests/XCTest-watchOS
+	url = ../../dfed/XCTest-watchOS.git
```
gfontenot: Should this be checked in?

README.md
```diff
@@ -6,7 +6,7 @@
 [![License](https://img.shields.io/cocoapods/l/Valet.svg)](http://cocoadocs.org/docsets/Valet)
 [![Platform](https://img.shields.io/cocoapods/p/Valet.svg)](http://cocoadocs.org/docsets/Valet)
 
-Valet lets you securely store data in the iOS, tvOS, or macOS Keychain without knowing a thing about how the Keychain works. It’s easy. We promise.
+Valet lets you securely store data in the iOS, tvOS, watchOS, or macOS Keychain without knowing a thing about how the Keychain works. It’s easy. We promise.
```
gfontenot: 🎉

Tests/MacTests.swift
```diff
@@ -22,74 +22,74 @@ import Foundation
...
-
-class ValetMacTests: XCTestCase
-{
-    // This test verifies that we are neutralizing the zero-day Mac OS X Access Control List vulnerability.
-    // Whitepaper: https://drive.google.com/file/d/0BxxXk1d3yyuZOFlsdkNMSGswSGs/view
-    // Square Corner blog post: https://corner.squareup.com/2015/06/valet-beats-the-ox-x-keychain-access-control-list-zero-day-vulnerability.html
-    @available (OSX 10.10, *)
-    func test_setStringForKey_neutralizesMacOSAccessControlListVuln()
+#if os(macOS)
+    class ValetMacTests: XCTestCase
```
gfontenot: Nit: What do you think about leaving this un-indented. Feels kinda weird to indent an entire class just because it's conditionally compiled.

Tests/SecureEnclaveTests.swift
```diff
@@ -126,10 +127,12 @@ class SecureEnclaveTests: XCTestCase
         let sharedAccessGroupIdentifier: Identifier
         #if os(iOS)
             sharedAccessGroupIdentifier = Identifier(nonEmpty: "com.squareup.Valet-iOS-Test-Host-App")!
-        #elseif os(OSX)
+        #elseif os(macOS)
             sharedAccessGroupIdentifier = Identifier(nonEmpty: "com.squareup.Valet-macOS-Test-Host-App")!
         #elseif os(tvOS)
             sharedAccessGroupIdentifier = Identifier(nonEmpty: "com.squareup.Valet-tvOS-Test-Host-App")!
+        #elseif os(watchOS)
+            sharedAccessGroupIdentifier = Identifier(nonEmpty: "com.squareup.ValetTouchIDTestApp.watchkitapp.watchkitextension")!
```
gfontenot: Nit: Can we standardize this name like the others?

Tests/SynchronizableTests.swift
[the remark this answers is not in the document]
gfontenot: `XCTAssertEqual` will ultimately give a better error message in the case of failure. I don't know enough about our debug descriptions to know how much that actually matters here.

.gitmodules
dfed: Having it checked in seems like the easiest thing. Note that it should be checked in under the `Tests/XCTest-watchOS` directory.

Tests/MacTests.swift
dfed: Hard agree that it's weird. Indentation was the default on Xcode 9 with ctrl-i. On Xcode 10 the indentation goes away, so I'll go with that :)

Tests/SecureEnclaveTests.swift
dfed: solid catch!

dfed: Okay, so, after a small investigation this is a rabbit hole I'd prefer to avoid (or fix in a later PR). I tied the watch test to the `ValetTouchIDTestApp`, and... the way WatchKit apps/extensions work every extension needs to share the same prefix. So unless I want to change _all_ the bundle IDs involved here... this is probably better worth skipping for now.

Tests/SynchronizableTests.swift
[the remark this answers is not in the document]
dfed: Yeah alright I'll fix this. I'll even add a failure description :)

.gitmodules
gfontenot: I guess I'm actually just unclear of what this submodule is.

Tests/SecureEnclaveTests.swift
gfontenot: 👍 seems reasonable

.gitmodules
dfed: This submodule is the code that does the `XCTest` on `watchOS`. It originally lived in this PR, but I broke it out into it's own repo/submodule since it isn't tightly coupled to `Valet`. Code is here: https://github.com/dfed/XCTest-watchOS

gfontenot: Ah, gotcha. Is this a local url? Or can you define relative remote urls like this? Assuming it _is_ remote, can we update this to point at the absolute url for the sake of clarity?

Valet.xcodeproj/project.pbxproj
```diff
@@ -1759,7 +1777,7 @@
 				CURRENT_PROJECT_VERSION = 1;
 				DEBUG_INFORMATION_FORMAT = dwarf;
 				DEFINES_MODULE = YES;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = 9XUJ7M53NG;
```
NickEntin: Does this need to specify `DEVELOPMENT_TEAM`/`CODE_SIGN_IDENTITY` here?  Probably better to leave these blank.

dfed: I've needed the dev team in the past to get unit tests to work in a signed environment (I think). I'll try removing. If it passes, then yeah, let's remove :)

dfed:
we-did-it-sw5z69

nemesis:
Congrats guys! 🎉

PR #132, opened 2018-03-28
Remove SinglePromptSecureEnclaveValet from tvOS target

dfed:
It turns out that `LocalAuthentication` isn't available on tvOS... despite Xcode showing me that it was 🙄. Building for tvOS device (rather than sim) failed due to our `import LocalAuthentication`.

This PR removes `SinglePromptSecureEnclaveValet` from the tvOS target entirely.

### DIFF
--- a/Tests/SinglePromptSecureEnclaveTests.swift
+++ b/Tests/SinglePromptSecureEnclaveTests.swift
@@ -23,11 +23,10 @@ import Foundation
 import XCTest
 
 
-@available(tvOS 10.0, *)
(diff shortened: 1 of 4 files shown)


Tests/SinglePromptSecureEnclaveTests.swift
```diff
@@ -263,15 +217,3 @@ class SinglePromptSecureEnclaveTests: XCTestCase
         }
     }
 }
-
-private class AvailabilityEnforcer {
```
EricMuller22: So long sweet prince

PR #133, opened 2018-03-28
Run carthage build in CI

dfed:
Unfortunately, I didn't realize that `LocalAuthentication` wasn't supported on tvOS until _after_ I'd merged #126. What turned me onto the oversight was Carthage failing to build.

This PR makes it so that we run Carthage in CI, so that this kind of issue doesn't happen again.

Built on top of #132

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,10 @@ language: objective-c
 osx_image: xcode9
 before_script:
     - bundle install
+    - brew outdated carthage || brew upgrade carthage
 script:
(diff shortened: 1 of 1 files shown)


EricMuller22:
Devil's advocate - merge and rebase #132 for extra safety/assurance?

dfed:
I can't merge this first since this PR will fail to build (since Carthage build is currently failing).

I'll rebase this guy on top of #132 though. I like your thinking.

PR #134, opened 2018-03-30
Create sharedAccessGroup Valet when requested

dfed:
🔥 

Note to anyone who adopted v3.0-3.1.1:
- If you shipped an app with a `Valet.sharedAccessGroupValet` or `Valet.iCloudSharedAccessGroupValet`, you'll need to take steps to not lose secure data in your next release.

If you released an app with a `Valet.sharedAccessGroupValet`, you'll need to run `Valet.sharedAccessGroupValet(with: yourIdentifier, accessibility: yourAccessibility).migrateObjects(from: Valet.valet(with: yourIdentifier, accessibility: yourAccessibility), removeOnCompletion: true)` before accessing data in your `Valet.sharedAccessGroupValet`

If you released an app with a `Valet.iCloudSharedAccessGroupValet`, you'll need to run `Valet.iCloudSharedAccessGroupValet(with: yourIdentifier, accessibility: yourAccessibility).migrateObjects(from: Valet.iCloudValet(with: yourIdentifier, accessibility: yourAccessibility), removeOnCompletion: true)` before accessing data in your `Valet.iCloudSharedAccessGroupValet`

### DIFF
--- a/Tests/SecureEnclaveTests.swift
+++ b/Tests/SecureEnclaveTests.swift
@@ -40,7 +40,18 @@ class SecureEnclaveTests: XCTestCase
         
         valet.removeObject(forKey: key)
     }
-    
+
+    // MARK: Initialization
+
+    func test_init_createsCorrectValet() {
+        let identifier = ValetTests.identifier
+        let accessControl = SecureEnclaveAccessControl.userPresence
+        XCTAssertEqual(SecureEnclaveValet.valet(with: identifier, accessControl: accessControl).service,
+                       Service.standard(identifier, .secureEnclave(accessControl)))
+        XCTAssertEqual(SecureEnclaveValet.sharedAccessGroupValet(with: identifier, accessControl: accessControl).service,
+                       Service.sharedAccessGroup(identifier, .secureEnclave(accessControl)))
+    }
+
     // MARK: Equality
     
     func test_secureEnclaveValetsWithEqualConfiguration_haveEqualPointers()
--- a/Tests/SecureEnclaveTests.swift
+++ b/Tests/SecureEnclaveTests.swift
@@ -43,13 +43,13 @@ class SecureEnclaveTests: XCTestCase
 
     // MARK: Initialization
 
-    func test_init_createsCorrectValet() {
+    func test_init_createsCorrectBackingService() {
         let identifier = ValetTests.identifier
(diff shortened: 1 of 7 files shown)


Valet.podspec
```diff
@@ -14,18 +14,4 @@ Pod::Spec.new do |s|
   s.macos.deployment_target = '10.11'
 
   s.tvos.exclude_files = 'Sources/SinglePromptSecureEnclaveValet.swift'
-
```
dfed: cc @dnkoutso. I'm pretty darned sure that the `test_spec` can't handle test environments that require the Keychain Sharing entitlement.

EricMuller22:
Failing on master:
screen shot 2018-03-30 at 2 30 29 pm

Passing on this branch:
screen shot 2018-03-30 at 2 31 26 pm

Tests/SecureEnclaveTests.swift
```diff
@@ -40,7 +40,18 @@ class SecureEnclaveTests: XCTestCase
...
+    // MARK: Initialization
+
+    func test_init_createsCorrectBackingService() {
+        let identifier = ValetTests.identifier
+
+        SecureEnclaveAccessControl.allValues().forEach { accessControl in
+            let backingService = SecureEnclaveValet.valet(with: identifier, accessControl: accessControl).service
+            XCTAssertEqual(backingService, Service.standard(identifier, .secureEnclave(accessControl)))
+        }
+    }
```
NickEntin: Shouldn't we also include shared access tests here (and in `SinglePromptSecureEnclaveTests`)?

Valet.podspec
NickEntin: Are there any side effects of removing the test spec?

dfed:
@NickEntin the only side-effect of removing the test spec is that publishing the spec will no longer run tests. Since we're running all these tests in CI, we're okay. The test_spec was only recently added anyways

Valet.podspec
dnkoutso: if its urgent you can remove this now and I can re-enable.

dfed: This is indeed urgent. Thank you!

dfed:
Note to anyone pursuing this PR: I've updated the release notes for 3.0.0-3.1.1 to note the known issue and workaround. I've also made a note in the release notes for 3.1.2 to call out the fix and workaround.

PR #135, opened 2018-03-30
Fix warnings on Xcode 9.3

dfed:
Warnings are bad, mkay?


Sources/Internal/Keychain.swift
```diff
@@ -202,11 +202,18 @@ internal final class Keychain {
                 return SecItem.DataResult.success(Set([singleKey]))
                 
             } else if let multipleMatches = collection as? [[String: AnyHashable]] {
+                #if swift(>=4.1)
```
NickEntin: Shouldn't this always build with Swift 4.0 based on the `.swift-version` file?

dfed: If only that's how `.swift-version` worked. All that does is let Cocoapods decide that two projects are on different swift major version and therefore won't work well together.

NickEntin: I thought it also affected the `SWIFT_VERSION` in the `Pods.xcodeproj` build settings?

dfed: It may. But if you use submodules, you'll get a build warning. So we should fix this :p

NickEntin: Ahh right, good point.

Issue #141, opened 2018-06-19
support xcode 9


wjmelements:
Valet uses swift 2 which xcode 9 does not support. After doing pod install with

```rb
    platform :ios, '9.0'
    use_frameworks!
    pod 'Valet'
```

I get 

```
The “Swift Language Version” (SWIFT_VERSION) build setting must be set to a supported value for targets which use Swift. This setting can be set in the build settings editor.
```

Swift language version can be specified per-target, so it would make sense to set it for this project.

dfed:
@wjmelements I think you may have a problem with your Pods configuration. Valet 3.0 was a complete rewrite from Objc -> Swift, and we rewrote it in [Swift 4](https://github.com/square/Valet/blob/master/.swift-version).

My guess is you'll get the above error importing any/all Swift projects. I'm closing this out.

If you're looking for a quick workaround that doesn't address the underlying configuration issue, you can try pinning to Valet 2.0, which was written in Objective C.

wjmelements:
Okay, similar to https://github.com/jessesquires/JSQCoreDataKit/issues/98, the work that needs to be done here is to specify the required swift version in the podfile.

> s.pod_target_xcconfig = { 'SWIFT_VERSION' => '4.0' }

PR #142, opened 2018-06-23
Swift Version

wjmelements:
This fixes https://github.com/square/Valet/issues/141 using the same strategy as https://github.com/jessesquires/JSQCoreDataKit/pull/99.
There is an error for non-swift projects that use the swift Valet.
It looks like this:
```
The “Swift Language Version” (SWIFT_VERSION) build setting must be set to a supported value for targets which use Swift. This setting can be set in the build settings editor.
```
Though the suggested workaround was to use Valet 2.0, I believe this fixes the underlying issue.
Reviewers @dfed

### DIFF
--- a/Valet.podspec
+++ b/Valet.podspec
@@ -6,12 +6,14 @@ Pod::Spec.new do |s|
   s.homepage = 'https://github.com/square/Valet'
   s.authors  = 'Square'
   s.source   = { :git => 'https://github.com/square/Valet.git', :tag => s.version }
+  s.pod_target_xcconfig = { 'SWIFT_VERSION' => '4.0' }
   s.source_files = 'Sources/**/*.{swift,h}'
   s.public_header_files = 'Sources/*.h'
   s.frameworks = 'Security'
   s.ios.deployment_target = '9.0'
   s.tvos.deployment_target = '9.0'
   s.macos.deployment_target = '10.11'
+  git@github.com:square/Valet.git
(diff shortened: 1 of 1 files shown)


wjmelements:
I have validated this change by pulling my podspec. I no longer see the error.

Valet.podspec
```diff
@@ -6,6 +6,7 @@ Pod::Spec.new do |s|
   s.homepage = 'https://github.com/square/Valet'
   s.authors  = 'Square'
   s.source   = { :git => 'https://github.com/square/Valet.git', :tag => s.version }
+  s.pod_target_xcconfig = { 'SWIFT_VERSION' => '4.0' }
```
dnkoutso: I wouldnt use that. There is an official DSL `s.swift_version` that you can use.

wjmelements:
I have now also verified the `swift_version` podspec.

dnkoutso:
change looks good the only thing to know is that this DSL exists since CocoaPods 1.4.0 so consumers earlier than that will not be using this still.

dfed:
Glad it looks good. Do we know what the recommended fix is for folks on prior Cocoapods versions?

I’ll merge this tomorrow when I’m at a computer + release a new version. Thank you for the fix!

wjmelements:
I suspect the prior version of this podspec (caf8217) will work for earlier pod versions.

PR #146, opened 2018-07-26
Add compatibility shim for flatMap rename

gfontenot:
In Swift 4.1, the version of `flatMap` with the shape
`[T] -> (T -> U?) -> [U]` has been renamed to `compactMap`. By adding this
compatibility shim for Swift versions < 4.1, we can use the new spelling
without losing any backwards compatibility on older Swift versions.

We're also not exporting this, so it has no affect on consuming code. However,
apps that build Valet as a part of their app with Swift 4.1+ will no longer
see warnings about `flatMap` being deprecated.

### DIFF
--- a/Sources/Internal/Keychain.swift
+++ b/Sources/Internal/Keychain.swift
@@ -202,17 +202,10 @@ internal final class Keychain {
                 return SecItem.DataResult.success(Set([singleKey]))
                 
             } else if let multipleMatches = collection as? [[String: AnyHashable]] {
-                #if swift(>=4.1)
                 return SecItem.DataResult.success(Set(multipleMatches.compactMap({ attributes in
                     let key = attributes[kSecAttrAccount as String] as? String
                     return key != canaryKey ? key : nil
                 })))
-                #else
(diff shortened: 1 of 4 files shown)


Sources/Internal/Keychain.swift
```diff
@@ -208,7 +208,7 @@ internal final class Keychain {
                     return key != canaryKey ? key : nil
                 })))
                 #else
-                return SecItem.DataResult.success(Set(multipleMatches.flatMap({ attributes in
+                return SecItem.DataResult.success(Set(multipleMatches.compactMap({ attributes in
```
dfed: I think we can get rid of the `#if swift(>=4.1)` above give the new shim.

gfontenot:
🤦‍♂️ We're on an old version, and I didn't even think to check. sorry. That being said, this might be a better long-term solution so that we remove the inline duplication. This would also let us use `compactMap` anywhere else we need to without needing to switch, and when we drop support for Swift 4.0- we can just delete this file from the project.

gfontenot:
Done! I'd be happy to merge and release, but I don't have push access to this repo. Do you have those superpowers?

PR #150, opened 2018-08-14
Use -hash instead of -hashValue.

### LINKED ISSUE #149: Hashable conformance in Swift 4.2
Running a project on Xcode 10, I see some warnings around the implementation of `hashValue` on a few classes:

`warning: override of 'NSObject.hashValue' is deprecated; override 'NSObject.hash' to get consistent hashing behavior`

This happens for the `SecureEnclaveValet`, `SinglePromptSecureEnclaveValet`, and `Valet` classes.

SlaunchaMan:
This improves compatibility with Swift 4.2. Fixes #149.

### DIFF
--- a/Sources/SecureEnclaveValet.swift
+++ b/Sources/SecureEnclaveValet.swift
@@ -90,8 +90,8 @@ public final class SecureEnclaveValet: NSObject {
     
     // MARK: Hashable
     
-    public override var hashValue: Int {
-        return service.description.hashValue
+    public override var hash: Int {
+        return service.description.hash
     }
     
     // MARK: Public Properties
--- a/Sources/SecureEnclaveValet.swift
+++ b/Sources/SecureEnclaveValet.swift
@@ -91,7 +91,7 @@ public final class SecureEnclaveValet: NSObject {
     // MARK: Hashable
     
     public override var hash: Int {
-        return service.description.hash
+        return service.description.hashValue
     }
     
     // MARK: Public Properties
--- a/Sources/SinglePromptSecureEnclaveValet.swift
+++ b/Sources/SinglePromptSecureEnclaveValet.swift
@@ -91,8 +91,8 @@ public final class SinglePromptSecureEnclaveValet: NSObject {
     
     // MARK: Hashable
     
-    public override var hashValue: Int {
-        return service.description.hashValue
+    public override var hash: Int {
+        return service.description.hash
     }
     
     // MARK: Public Properties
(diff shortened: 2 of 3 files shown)


Sources/SecureEnclaveValet.swift
```diff
@@ -90,8 +90,8 @@ public final class SecureEnclaveValet: NSObject {
     
     // MARK: Hashable
     
-    public override var hashValue: Int {
-        return service.description.hashValue
+    public override var hash: Int {
```
dfed: Can we write this as
```
#if swift(>=4.2)
public override var hash: Int {
    return service.description.hash
}
#else
public override var hashValue: Int {
    return service.description.hashValue
}
#endif
```

Here and below? That way I can release a version with this sha without breaking Xcode 9 folks.

SlaunchaMan: I thought about that at first, too, but this approach still produces the warning in Xcode 10. It's not that `hash` is a new API; rather, `hashValue` comes from the `Hashable` protocol and `hash` comes from `NSObject`; since these classes inherit from `NSObject` the guidance from Xcode is to use `hash` instead. I did test that this code works on Xcode 9.3.

dfed: Got it! Appreciate the explanation.

Your explanation helped me find some supportive documentation in [Swift's CHANGELOG](https://github.com/apple/swift/blob/master/CHANGELOG.md) re using `hash` in `NSObject` subclasses. Under the subheading of [SE-0206](https://github.com/apple/swift-evolution/blob/master/proposals/0206-hashable-enhancements.md) under Swift 4.2, the log mentions that the new `hashValue`/`Hashable` changes `don't affect Foundation's hashing interface. Classes that subclass NSObject should override the hash property, like before.`

With that, your PR LGTM. Will merge.

Sources/SecureEnclaveValet.swift
```diff
@@ -90,8 +90,8 @@ public final class SecureEnclaveValet: NSObject {
...
+        return service.description.hash
```
dfed: Actually, one more question: do we want to use `hash` on the `description`, or `hashValue`? Or does it not matter because `String` and `NSString` are bridged due to our `Foundation` import?

SlaunchaMan: Good catch! I’ll revert these back to use `hashValue` since `description` is defined as a Swift `String`.

Sources/SecureEnclaveValet.swift
SlaunchaMan: Thanks for finding that. I was trying to find some actual concrete documentation on `hash` vs. `hashValue` and this looks like the closest thing.

SlaunchaMan:
@dfed Any chance we can get a patch release on CocoaPods?

dfed:
Doing this now. Thanks for the nudge!

Issue #143, opened 2018-06-28
Differentiate between errSecUserCanceled and errSecAuthFailed


umbri:
Thank you for sharing this.

I think that ```errSecUserCanceled``` and ```errSecAuthFailed``` are 2 different important errors and grouping them in one like ```userCancelled``` is not the right way, my propose is to split them in ```userCancelled``` and ```authFailed```, ref: #144 

Thank you

dfed:
@umbri thanks for the issue and pull request!

We currently conflate these errors because both `authFailed` and` userCancelled` are triggered by events that effectively mean that the user cancelled. `authFailed` is returned when the user failed to input their password or locked the screen while you were prompting for their password – I wouldn't want our library's customers to have to think about the difference between these a user failing to enter a password and hitting cancel, since they are effectively the same thing. When you get the `userCancelled` return, you know there is valid data in the keychain, but that the user didn't let you access it.

https://github.com/square/Valet/pull/73#issuecomment-184892249 is where we originally discussed treating both `errSecUserCanceled` and `errSecAuthFailed` as `userCancelled`. If you disagree with that logic, it'd be great to hear why. I'd also be curious what use-cases having separate errors for `authFailed` vs` userCancelled` will enable.

dfed:
Got it. Interesting! Given that `errSecAuthFailed` can be returned when a user accidentally hits the power button, I'd be careful deleting all data when you get that error code.

I need to sit on this use case a bit. I'm not a huge fan of exposing authorization failure as a separate `case` from `userCancelled`, since in almost all cases they're the same. I'm thinking about maybe adding an additional flag as associated data on `userCancelled` (i.e. `case userCancelled(dueToAuthorizationFailure: Bool)`. Adding an associated value would be a non-breaking change for existing clients who use `switch` statements without a `default`, and allows us to keep the simple flow for the majority use case. I like the associated data approach better, but I'm still not convinced that this is something that our library wants to expose – a guiding tenant of our library is to make the keychain exceptionally simple to use, which means limiting API surface.

@EricMuller22, any thoughts?

umbri:
@dfed I like ```userCancelled(dueToAuthorizationFailure: Bool)``` too

gfontenot:
I haven't been involved in this project for long, so my opinions here shouldn't necessarily take precedence over other people's, but:

[...]

PR #152, opened 2018-10-02
Differentiate between UserCancelled and AuthFailed

dfed:
Addresses #143. I'm not convinced we want to merge this, but at the very least this is a good straw man.

Thoughts @umbri, @EricMuller22, @NickEntin, and @gfontenot?

### DIFF
--- a/Sources/SecureEnclave.swift
+++ b/Sources/SecureEnclave.swift
@@ -99,22 +99,9 @@ public final class SecureEnclave {
             secItemQuery[kSecUseOperationPrompt as String] = userPrompt
         }
         
-        switch Keychain.object(forKey: key, options: secItemQuery) {
-        case let .success(data):
-            return .success(data)
-            
-        case let .error(status):
-            let authFailed = (status == errSecAuthFailed)
-            let userCancelled = (status == errSecUserCanceled)
-            if userCancelled || authFailed {
-                // While user cancelling is self-explanatory, authorization failure can be triggered by a user entering the wrong password or locking the device while a secure enclave prompt is on screen. While we recommend applications treat these two failure types as a user cancel, some applications want to differentiate between users bailing out of the flow and failed password attempts. For more discussion on the differences between these statuses, see https://github.com/square/Valet/pull/73 and https://github.com/square/Valet/issues/143.
-                return .userCancelled(dueToAuthorizationFailure: authFailed)
-            } else {
-                return .itemNotFound
-            }
-        }
+        return Keychain.object(forKey: key, options: secItemQuery).asSecureEnclaveResult
     }
-    
+
     /// - parameter key: The key to look up in the keychain.
     /// - parameter options: A base query used to scope the calls in the keychain.
     /// - returns: `true` if a value has been set for the given key, `false` otherwise.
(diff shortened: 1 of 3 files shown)


Sources/SecureEnclave.swift
```diff
@@ -163,9 +165,11 @@ public final class SecureEnclave {
             return .success(string)
             
         case let .error(status):
-            let userCancelled = (status == errSecUserCanceled || status == errSecAuthFailed)
-            if userCancelled {
-                return .userCancelled
+            let authFailed = (status == errSecAuthFailed)
+            let userCancelled = (status == errSecUserCanceled)
+            if userCancelled || authFailed {
```
gfontenot: Is there a reasonable way to consolidate this logic with the duplicate logic on 110? Feels like this might be a bit fragile over time.

Sources/SecureEnclave.swift
```diff
@@ -29,8 +29,8 @@ public final class SecureEnclave {
     public enum Result<Type: Equatable>: Equatable {
         /// Data was retrieved from the keychain.
         case success(Type)
-        /// User dismissed the user-presence prompt.
-        case userCancelled
+        /// User dismissed the user-presence prompt. If `dueToAuthorizationFailure` is true, the user either entered the wrong password or locked the device while the prompt was presented.
+        case userCancelled(dueToAuthorizationFailure: Bool)
```
gfontenot: Is it worth breaking this out into a new case? Feels a bit odd to tack this on to the canceled state to me.

dfed: It's worth reading up on https://github.com/square/Valet/pull/73 and https://github.com/square/Valet/issues/143. UserCancelled and AuthFailure have a good bit of overlap, and I don't think we should make our consumers tell the difference between them.

Specifically, AuthFailure can mean that the user locked their screen while the prompt was up, which feels a lot like a UserCancel even to me. We've been conflating UserCancelled and AuthFailure since we introduced the ability to detect UserCancel back in https://github.com/square/Valet/pull/73. What we're doing in this PR is giving our consumers the ability to tease these cases apart if they feel like it, without breaking existing adopters, and without adding burden to new adopters who don't care about the differences between these cases.

I think it's important to view the consumer-facing API without a whole lot of knowledge of how the underlying keychain works. There are currently 3 options in this enum, which I believe accurately define what happens when you request a value from the Secure Enclave:
1. `success` – i.e. "We got the value!"
2. `userCancelled` - i.e. "The user bailed out of the prompt somehow"
3. `itemNotFound` - i.e. "There is no item in the keychain"

In this world, authorization failure (i.e. "incorrect passcode" or "user locked the screen") feels like a subset of `userCancelled`. If I could tease "incorrect passcode" apart from "user locked the screen", I'd add another enum case. But since I can't, and I can't provide more insight than "we got an AuthFailed response", I'd rather keep auth failure as an associated value on the general concept of `userCancelled`. Does that make sense?

gfontenot: 👍

dfed: So, the real question becomes: do we want to add this API? I feel a little weird exposing this kind of keychain-specific API; it violates some of our promise of "without knowing a thing about how the Keychain works". But, it could also be useful information per the request in #143. This API errs on the side of enabling our consumers to do what they need with our API, rather than keeping the API as simple as possible. But I think we may have struck a reasonable balance between simplicity for the average consumer and making complex use cases possible.

Does this API pass the smell test for you? Does it seem like we've struck a good balance / made a reasonable trade-off here?

gfontenot: Added my higher level comment on #143

dfed:
closing per 
https://github.com/square/Valet/issues/143#issuecomment-426405483

Issue #137, opened 2018-04-12
Migrating non kSecClassGenericPassword item results in data loss


MarkVillacampa:
Since all items stored by Valet are of type `kSecClassGenericPassword`, wouldn't migrating an item of a different type like `kSecClassInternetPassword` loose attribute information like `kSecAttrServer`?

dfed:
Good point! Maybe we could add a requirement that `query` provided to `migrateObjects` had the `kSecClass as String : kSecClassGenericPassword as String` as a key:value pair. Seems like a good safety net.

In the meantime though, sending a query to migrate values that are not `kSecClassGenericPassword` is considered user error. I'm going to adjust the title of this issue accordingly, and label this an enhancement request.

If you'd like to submit a PR to `ErrorHandler.assert` that the `migrateObjects` `query` has `kSecClassGenericPassword`, we'd accept it. Just make sure to add a unit test or two :)

PR #153, opened 2018-10-02
Enforce that migration specifies kSecClassGenericPassword to avoid data loss

dfed:
Addresses #137

cc @MarkVillacampa

### DIFF
--- a/Sources/Internal/Keychain.swift
+++ b/Sources/Internal/Keychain.swift
@@ -250,7 +250,7 @@ internal final class Keychain {
             return .invalidQuery
         }
         
-        guard query[kSecClass as String] as? String == kSecClassGenericPassword as String else {
+        guard query[kSecClass as String] as? String as CFString? == kSecClassGenericPassword else {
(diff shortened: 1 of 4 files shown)


Sources/Internal/Keychain.swift
```diff
@@ -250,8 +250,8 @@ internal final class Keychain {
             return .invalidQuery
         }
         
-        guard let _ = query[kSecClass as String] as? String else {
-            ErrorHandler.assertionFailure("Migration requires a kSecClass to be set to a valid kSecClass string.")
+        guard query[kSecClass as String] as? String == kSecClassGenericPassword as String else {
```
gfontenot: I assume `kSecClassGenericPassword` is an `NSString`? Is it worth casting the dictionary result `NSString `to avoid needing to cast `kSecClassGenericPassword` at all?

dfed: It's a `CFString`, which complicates casts a bit. However, your q made me realize we can make this a little less insane. New commit incoming.

Issue #155, opened 2018-10-02
XCTest-watchOS.git does not exist Error 


elyasns:
On the new release `v3.2.0`, there is an error in Carthage build for watchOS test git repo.

`fatal: repository '../../dfed/XCTest-watchOS.git' does not exist`

dfed:
Oh no. Thanks for the report! I’m guessing that this is because I used a relative git  submodules path. I’ll see if I can fix this later today. My guess is I need to move to using HTTPS.

dfed:
Release 3.2.1 should fix this.

PR #156, opened 2018-10-02
Use https for git submodule

dfed:
This change fixes Carthage. While it's unfortunate to not allow git cloning of submodules to reflect the method of the original clone (i.e. cloning Valet via `ssh` will now clone `XCTest-watchOS` via `https`), it's better to make sure our distribution mechanisms are not broken.

I'm filing an issue with Carthage to see if they can fix the underlying issue. But for now, I'm putting out a fix.

cc @elyasns


dfed:
Carthage issue has been filed: https://github.com/Carthage/Carthage/issues/2600

dfed:
Note that this PR addresses #155

PR #158, opened 2018-10-02
Unify repeated logic

dfed:
Takes the learning from https://github.com/square/Valet/pull/152#discussion_r222036403 and applies it, even though we nixed #152

### DIFF
--- a/Sources/SecureEnclave.swift
+++ b/Sources/SecureEnclave.swift
@@ -33,7 +33,24 @@ public final class SecureEnclave {
         case userCancelled
         /// No data was found for the requested key.
         case itemNotFound
-        
+
+        // MARK: Initialization
+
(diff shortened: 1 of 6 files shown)


Sources/SecureEnclave.swift
```diff
@@ -157,11 +146,20 @@ public final class SecureEnclave {
...
-            return .success(string)
-            
+
+        return Keychain.string(forKey: key, options: secItemQuery).asSecureEnclaveResult
+    }
+}
+
+// MARK: DataResult Extension
+
+extension SecItem.DataResult where SuccessType: Equatable {
```
gfontenot: What do you think about making this an initializer on `SecureEnclave.Result` instead? We can still use an extension to add the `Equatable` restriction, but it's a bit more in line with common Swift design guidelines.

PR #161, opened 2018-10-24
Attempt to run per-platform builds in parallel

dfed:
If it builds, let's merge it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,4 @@ before_script:
   - bundle exec pod lib lint --verbose --fail-fast
   - carthage build --verbose --no-skip-current
 script:
-  - if [ -n "$DESTINATION" ]
-  then xcodebuild -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -destination "$DESTINATION" -configuration Debug -PBXBuildsContinueAfterErrors=0 $ACTION
-  else xcodebuild -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -configuration Debug -PBXBuildsContinueAfterErrors=0 $ACTION
-  fi
+  - ./Scripts/ci.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,7 @@ before_script:
   - bundle exec pod lib lint --verbose --fail-fast
   - carthage build --verbose --no-skip-current
 script:
-  - if [ -n "$DESTINATION" ]; then xcodebuild -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -destination "$DESTINATION" -configuration Debug -PBXBuildsContinueAfterErrors=0 $ACTION; else xcodebuild -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -configuration Debug -PBXBuildsContinueAfterErrors=0 $ACTION; fi
+  - if [ -n "$DESTINATION" ]
+  then xcodebuild -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -destination "$DESTINATION" -configuration Debug -PBXBuildsContinueAfterErrors=0 $ACTION
+  else xcodebuild -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -configuration Debug -PBXBuildsContinueAfterErrors=0 $ACTION
+  fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,15 @@
 language: objective-c
 osx_image: xcode9
+env:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,12 +1,15 @@
...
 script:
-    - xcodebuild -project Valet.xcodeproj -scheme "Valet iOS" -sdk iphonesimulator -destination "platform=iOS Simulator,OS=11.0,name=iPhone 7" -configuration Debug -PBXBuildsContinueAfterErrors=0 build test
-    - xcodebuild -project Valet.xcodeproj -scheme "Valet tvOS" -sdk appletvsimulator11.0 -destination "platform=tvOS Simulator,name=Apple TV" -configuration Debug -PBXBuildsContinueAfterErrors=0 build test
-    - xcodebuild -project Valet.xcodeproj -scheme "Valet Mac" -sdk macosx10.13 -configuration Debug -destination "platform=OS X" -PBXBuildsContinueAfterErrors=0 build test
-    - xcodebuild -project Valet.xcodeproj -scheme "Valet watchOS" -sdk watchos4.0 -configuration Debug -PBXBuildsContinueAfterErrors=0 build
-    - bundle exec pod lib lint --verbose --fail-fast
-    - carthage build --verbose --no-skip-current
+  - if [ -n "$DESTINATION" ]; then xcodebuild -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -destination "$DESTINATION" -configuration Debug -PBXBuildsContinueAfterErrors=0 $ACTION; else xcodebuild -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -configuration Debug -PBXBuildsContinueAfterErrors=0 $ACTION; fi
```
dfed: This line is a bit gross: we can't specify a watch destination in CI, so we need to omit the `-destination` argument if one hasn't been set. This approach clearly doesn't scale to multiple parameters being optional. If someone with better bash skills can think of a better way to do this, I'm all ears.

.travis.yml
```diff
@@ -1,12 +1,18 @@
 language: objective-c
 osx_image: xcode9
+env:
+  - SCHEME="Valet iOS";SDK="iphonesimulator";DESTINATION="platform=iOS Simulator,OS=11.0,name=iPhone 7";ACTION="build test"
+  - SCHEME="Valet tvOS";SDK="appletvsimulator11.0";DESTINATION="platform=tvOS Simulator,name=Apple TV";ACTION="build test"
+  - SCHEME="Valet Mac";SDK="macosx10.13";DESTINATION="platform=OS X";ACTION="build test"
+  - SCHEME="Valet watchOS";SDK="watchos4.0";DESTINATION="";ACTION="build"
```
gfontenot: How does this work? I'm assuming this is a travis feature? Is this spinning up new builds for each env?

.travis.yml
```diff
@@ -1,12 +1,18 @@
...
-    - xcodebuild -project Valet.xcodeproj -scheme "Valet iOS" -sdk iphonesimulator -destination "platform=iOS Simulator,OS=11.0,name=iPhone 7" -configuration Debug -PBXBuildsContinueAfterErrors=0 build test
-    - xcodebuild -project Valet.xcodeproj -scheme "Valet tvOS" -sdk appletvsimulator11.0 -destination "platform=tvOS Simulator,name=Apple TV" -configuration Debug -PBXBuildsContinueAfterErrors=0 build test
-    - xcodebuild -project Valet.xcodeproj -scheme "Valet Mac" -sdk macosx10.13 -configuration Debug -destination "platform=OS X" -PBXBuildsContinueAfterErrors=0 build test
-    - xcodebuild -project Valet.xcodeproj -scheme "Valet watchOS" -sdk watchos4.0 -configuration Debug -PBXBuildsContinueAfterErrors=0 build
-    - bundle exec pod lib lint --verbose --fail-fast
-    - carthage build --verbose --no-skip-current
+  - if [ -n "$DESTINATION" ]
+  then xcodebuild -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -destination "$DESTINATION" -configuration Debug -PBXBuildsContinueAfterErrors=0 $ACTION
+  else xcodebuild -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -configuration Debug -PBXBuildsContinueAfterErrors=0 $ACTION
+  fi
```
gfontenot: Is it worth splitting this out into a standalone shell script to avoid the wacky formatting?

.travis.yml
dfed: This is entirely a travis feature – I just learned about it yesterday. Documentation is here: https://docs.travis-ci.com/user/speeding-up-the-build/

.travis.yml
dfed: That would probably be way better. Will attempt that before merging.

.travis.yml
dfed: Basically we're setting environment variables that will be given to each parallel instance.

PR #165, opened 2019-01-07
Make Accessibility and CloudAccessibility conform to CaseIterable ...

louisdebaere:
... removing the need for an allValues property


Sources/Accessibility.swift
```diff
@@ -84,18 +84,4 @@ public enum Accessibility: Int, CustomStringConvertible, Equatable {
         
         return accessibilityAttribute as String
     }
-    
-    // MARK: Internal
-    
-    internal static func allValues() -> [Accessibility] {
```
dfed: In order to maintain backwards compatibility with older Swift versions, we'll need to keep this method. However, we can deprecate it with a message that this property will be removed in a later version. Something like:
```
@available(swift, obsoleted: 4.2, renamed: "allCases", message: "This method should be removed in the next major version bump of Valet")
```

Technically we don't need to use deprecation syntax since this property is `internal`, but it can't hurt to use built-in Swift syntax to make our intent explicit.

Sources/Accessibility.swift
```diff
@@ -22,7 +22,7 @@ import Foundation
 
 
 @objc(VALAccessibility)
-public enum Accessibility: Int, CustomStringConvertible, Equatable {
+public enum Accessibility: Int, CustomStringConvertible, Equatable, CaseIterable {
```
dfed: This is exciting! However, adopting this new Swift API will break Xcode 9 consumers, or any consumer still using a Swift version prior to 4.2. I'd rather not do a major version bump in order to support `CaseIterable`.

In order to land this, we'll need to maintain backwards compatibility. We can do so by adding the following to our SwiftCompatibility.swift file:
```
#if !swift(>=4.2)

public protocol CaseIterable {

  associatedtype AllCases : Collection where Self.AllCases.Element == Self

  static var allCases: Self.AllCases { get }
}

#endif
```

We'll then need to add the following to `Accessibility`:

```
    #if !swift(>=4.2)
    public static var allCases: [Accessibility] {
        return allValues()
    }
    #endif
```

Similarly, we'll need to add the following to CloudAccessibility:
```
    #if !swift(>=4.2)
    public static var allCases: [CloudAccessibility] {
        return allValues()
    }
    #endif
```

Sources/CloudAccessibility.swift
```diff
@@ -52,14 +52,4 @@ public enum CloudAccessibility: Int, CustomStringConvertible, Equatable {
     public var secAccessibilityAttribute: String {
         return accessibility.secAccessibilityAttribute
     }
-    
-    // MARK: Internal
-    
-    internal static func allValues() -> [CloudAccessibility] {
```
dfed: Same comment here as above. Let's keep this property and add an appropriate `@available` tag.

Sources/Accessibility.swift
dfed: Something to note about this change is that the ability to enumerate all values will now be a `public` feature rather than an `internal` feature. This means that adding a new case to these enums may constitute a breaking change, which is not an ideal state.

Given that `allValues()` is currently only used by our test code, maybe we should instead make `Accessibility` and `CloudAccessibility` conform to `CaseIterable` via an extension in the Test target? Note that we'd still need the above backwards compatibility code to allow us to continue running Valet's CI on the oldest supported Xcode version.

louisdebaere: Sadly implementation of `CaseIterable` cannot be automatically synthesized in an extension in a different file to the type

louisdebaere: Implementing it manually wouldn't be an improvement over the current `allValues` function

dfed: > Sadly implementation of CaseIterable cannot be automatically synthesized in an extension in a different file to the type

Today I learned! That is indeed unfortunate.

> Implementing it manually wouldn't be an improvement over the current allValues function
 
I completely agree. Maybe we should leave this code as-is then? Or do you think we should make `allCases` public? I'm open to hearing why `allCases` could be useful to external developers, but I'll admit I'm skeptical.

louisdebaere: > I completely agree. Maybe we should leave this code as-is then?
Yes, agreed
> Or do you think we should make allCases public? I'm open to hearing why allCases could be useful to external developers, but I'll admit I'm skeptical.
I don't, I admit I did not consider the impact of this change at first. Thanks for pointing it out!

dfed: No worries at all! Thank you for being responsive to feedback, and for exploring what our repo would look like with `CaseIterable`!

Given the conclusions in the above discussion, I'm going to close this PR. Thank you again for helping keep Valet up-to-date, and I look forward to more contributions from you in the future 😄

PR #169, opened 2019-03-28
Xcode 10.2 cleanup

dfed:
This PR does a few things:
1. Changes localization structure to what Xcode 10.2 recommends. Given that we have no localizations I think this shouldn't affect anything.
1. Runs the Swift 5 migrator. There are no source changes required for us to update to Swift 5.
1. Shares the `Valet watchOS Test Host App Extension` scheme so folks can easily run the watch tests locally
1. Makes project changes recommended by Xcode 10.2

Happy to break this out into separate PRs, but they're simple enough. I don't think we need a version bump with this change – curious if CI agrees.


dfed:
Looks like updating the `SWIFT_VERSION` causes issues 😬. Will revert that part.

PR #170, opened 2019-04-12
Update Version To Swift 5

jakeholland:
Updates project to Swift 5 and updates Readme. Remove old Swift 4.0 support.


dfed:
We can't remove Swift 4 support without a major version bump, since removing Swift 4 support is an [incompatible API change](https://semver.org).

I'd rather keep support for both Swift 4 and Swift 5 until we have some breaking changes we want to make. Given that there are [no source changes required](https://github.com/square/Valet/pull/169) to get Valet compile with Swift 5, what is the need for this change?

jakeholland:
@dfed You are definitely right about the major bump needed, that was my bad.

As far as the _need_ for this change. My main concern is that installing with Cocoapods still makes Valet build for Swift 4. 

Ideally there would be a way to specify multiple support Swift versions. I don't know of any way to do that currently, do you?

dfed:
@jakeholland that makes sense. The good news is that multi-swift-version compatibility [is coming to Cocoapods 1.7](https://github.com/CocoaPods/CocoaPods/issues/8191). The bad news is it isn't GM yet.

In the meantime, you can use a [post_install](https://guides.cocoapods.org/syntax/podfile.html#post_install) hook [like this](https://github.com/CocoaPods/CocoaPods/issues/6791#issuecomment-427042784) to set the Swift version on Pods as you integrate them.

jakeholland:
@dfed - I was just looking at that, haha. Yeah, that might be the only sulution as of now to keep compatibility for 4.X. This can be closed.

PR #173, opened 2019-05-18
Add support for new-style SecAccessControlCreateFlags constants

### LINKED ISSUE #172: Deprecation warnings with Xcode 10.2 / Swift 5 / iOS 12
I'm getting some deprecation warnings when adding Valet as a pod dependency to my project.

```text
Valet/Sources/SecureEnclaveAccessControl.swift:79:25: warning: 'touchIDAny' was deprecated in iOS 11.3: renamed to 'SecAccessControlCreateFlags.biometryAny'
                return .touchIDAny
                        ^
Valet/Sources/SecureEnclaveAccessControl.swift:79:25: note: use 'SecAccessControlCreateFlags.biometryAny' instead
                return .touchIDAny
                        ^~~~~~~~~~
                        SecAccessControlCreateFlags.biometryAny
Valet/Sources/SecureEnclaveAccessControl.swift:86:25: warning: 'touchIDCurrentSet' was deprecated in iOS 11.3: renamed to 'SecAccessControlCreateFlags.biometryCurrentSet'
                return .touchIDCurrentSet
                        ^
Valet/Sources/SecureEnclaveAccessControl.swift:86:25: note: use 'SecAccessControlCreateFlags.biometryCurrentSet' instead
                return .touchIDCurrentSet
                        ^~~~~~~~~~~~~~~~~
                        SecAccessControlCreateFlags.biometryCurrentSet
```

I think we should be able to safe-guard us for the future, and make the warning go away, by using `if #available(iOS 11.3, *)` 🤔 

Hopefully, I'll have time to look into a PR later today ☺️

LinusU:
Fixes #172

### DIFF
--- a/Sources/SecureEnclaveAccessControl.swift
+++ b/Sources/SecureEnclaveAccessControl.swift
@@ -76,14 +76,14 @@ public enum SecureEnclaveAccessControl: Int, CustomStringConvertible, Equatable 
             return .userPresence
         case .biometricAny:
             if #available(macOS 10.12.1, *) {
-                return .touchIDAny
+                return .init(rawValue: 2) // .biometryAny with Xcode 9 compatibility.
             } else {
                 ErrorHandler.assertionFailure(".biometricAny requires macOS 10.12.1.")
                 return .userPresence
             }
         case .biometricCurrentSet:
             if #available(macOS 10.12.1, *) {
-                return .touchIDCurrentSet
+                return .init(rawValue: 8) // .biometryCurrentSet with Xcode 9 compatibility.
             } else {
                 ErrorHandler.assertionFailure(".biometricCurrentSet requires macOS 10.12.1.")
                 return .userPresence


dfed:
🤔. Looks like Xcode 9 (our minimum-supported Xcode version) doesn't like the new fields. Looks like there is no good way to handle renames of option set values over multiple Xcode version.

You might want to consider (and I know this sucks) using `SecAccessControlCreateFlags.init(rawValue:)` to avoid the warning. Consider instead of
```
if #available(macOS 10.12.1, *) {
    return .touchIDAny
} else {
    ErrorHandler.assertionFailure(".biometricAny requires macOS 10.12.1.")
    return .userPresence
}
```

We could write
```
if #available(macOS 10.12.1, *) {
    return .init(rawValue: 2) // .biometryAny with Xcode 9 compatibility.
} else {
    ErrorHandler.assertionFailure(".biometricAny requires macOS 10.12.1.")
    return .userPresence
}
```

`biometryCurrentSet`'s raw value is `8`, so you could do the same thing there. 

Thoughts? I want to provide a warning-free Valet for you, but with Swift's limitations (and my not wanting to do a major version bump to drop Xcode 9), I'm not seeing better options.

dfed:
@LinusU let me know what you think of the above. I'd love to see a fix for our warnings land on `master` with you as the author.

That said, I don't want to wait forever on this – warnings are annoying, and I don't want to burden our consumers with them for longer than I must. If I don't hear from you by Monday June 3rd, I'll likely put up a PR to remove these warnings.

LinusU:
So sorry for the delay on this!

Took your approach and also rebased on master, thanks! ☺️

PR #174, opened 2019-05-24
Add Multiple Swift Versions Support

jakeholland:
With Cocoapods 1.7.0 we are now able to specify support for multiple swift versions at once. This PR bumps the minimum Cocoapods version to 1.7.0 and adds explicit support for Swift versions 4.0, 4.1, 4.2, and 5.0. Valet's version has also been bumped to 3.2.4.


Valet.podspec
```diff
@@ -1,12 +1,13 @@
 Pod::Spec.new do |s|
   s.name     = 'Valet'
-  s.version  = '3.2.3'
+  s.version  = '3.2.4'
+  s.cocoapods_version = '>= 1.7.0'
```
dfed: What does this line entail? Is this change backwards compatible?

jakeholland:
Looks like CI is using "Bundler version 1.15.4" and the bundle file is using Bundler version 2. Checking into it.

dnkoutso:
Its probably because https://github.com/square/Valet/pull/174/files#diff-e79a60dc6b85309ae70a6ea8261eaf95R76 in this PR

dfed:
@jakeholland re Bundler, you should be able to run `gem install bundler -v 1.16.1` if you have that version of `bundler` installed. I think it'll pass CI if you do that.

Valet.podspec
dnkoutso: It is I was just about to comment to that. It should be backwards compatible all the way back to CocoaPods 1.4.0 when the `swift_version` (singular) DSL was first introduced.

dfed: Awesome!

dnkoutso:
Hm saw the build failure. I will debug.

dnkoutso:
Oh its because of Xcode 9

dnkoutso:
Need to bump that. `lint` will always use the latest version of Swift supported unless you specify `--swift-version` in the command as a parameter

jakeholland:
@dnkoutso - Would you rather I specify the Xcode version, or bump the Xcode version in the `.travis.yml`?

dnkoutso:
If we say 5.0 is the latest supported version our default should `lint` against that so I recommend bumping...@dfed you OK with that?

Additionally we can setup multiple jobs one per Swift version we support and use `--swift-version` to ensure `lint` passes for all the versions we claim to support. This was as designed by the Swift versions DSL feature.

jakeholland:
@dnkoutso - I am cool with either option, bumping is obviously the easiest. But, if we update the jobs to include all supported Swift versions that would be the safest option. 🤔

dnkoutso:
You might need to be on Xcode 10.2 which was the first Xcode to support Swift 5.

jakeholland:
> You might need to be on Xcode 10.2 which was the first Xcode to support Swift 5.

Good call!

Valet.podspec
dnkoutso: You can remove this line btw. Once this PR lands prior versions all the way down to 1.4.0 will pick Swift 5.0 as their version.

dfed:
The downside of bumping: we no longer have a CI check on our lowest-supported Xcode version, which can lead to potential compiler issues. See how CI caught an Xcode 9 build failure here: https://github.com/square/Valet/pull/173

Linting on the swift version supported by CI seems fine. Alternatively, we could try running CI over multiple machines, but that seems out of scope for this PR. I'd prefer for us to continue to use the minimum supported Xcode version in CI if at all possible.

dfed:
If we get https://github.com/square/Valet/pull/175 working (it's looking promising), we can rebase this PR on top of `master` and lint on multiple swift/Xcode versions pretty easily.

dnkoutso:
If #175 lands then we can just update this PR to pass `--swift-version` for the Xcode 9 build.

dfed:
It’s merged! Please rebase. I made it easy for you to add the Swift version flag in CI too! Just change [this line](https://github.com/square/Valet/blob/e09d34ee04cb081e2807acade8994e3cd39923f9/.travis.yml#L18).

Valet.xcodeproj/project.pbxproj
```diff
@@ -1738,6 +1738,7 @@
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
 				TARGETED_DEVICE_FAMILY = 3;
 				TEST_HOST = "$(BUILT_PRODUCTS_DIR)/Valet tvOS Test Host App.app/Valet tvOS Test Host App";
+				TVOS_DEPLOYMENT_TARGET = 12.2;
```
dfed: I think we can remove the changes to this file.

jakeholland:
Closing in favor of #178 Since this has a bunch of unneeded commits

PR #175, opened 2019-05-24
Get CI running on Xcode 9 and Xcode 10.2

dfed:
Not sure if this will work. Inspired by [groue/GRDB.swift](https://github.com/groue/GRDB.swift/blob/master/.travis.yml)'s `.travis.yml` file.

Might help with https://github.com/square/Valet/pull/174


dfed:
I can't merge this if it goes green. It's running a single job per Xcode image, not 4.

dfed:
image
We're now spinning up the right number of jobs. @dnkoutso mind taking another look?

Scripts/ci.sh
```diff
@@ -1,8 +1,18 @@
...
 
-if [ -n "$DESTINATION" ]; then
-  xcodebuild -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -destination "$DESTINATION" -configuration Debug -PBXBuildsContinueAfterErrors=0 $ACTION
-else
-  xcodebuild -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -configuration Debug -PBXBuildsContinueAfterErrors=0 $ACTION
+if [ $ACTION == "xcode" ]; then
+  if [ -n "$DESTINATION" ]; then
+    xcodebuild -UseModernBuildSystem=NO -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -destination "$DESTINATION" -configuration Debug -PBXBuildsContinueAfterErrors=0 $XCODE_ACTION
+  else
+    xcodebuild -UseModernBuildSystem=NO -project Valet.xcodeproj -scheme "$SCHEME" -sdk $SDK -configuration Debug -PBXBuildsContinueAfterErrors=0 $XCODE_ACTION
```
dfed: Note that we need to skip using the modern build system because the modern build system builds our custom `XCTest` library for non-watch targets for some unknown reason. Once the `XCTest` target is in derived data, the new build system uses it instead of the system's `XCTest`. Using the old build system fixes the issue.

Gemfile
```diff
@@ -1,3 +1,3 @@
 source 'https://rubygems.org' do
-  gem 'cocoapods', '~> 1.4'
+  gem 'cocoapods', '~> 1.6'
```
NickEntin: Should this be 1.7?  They just released 1.7.0

dfed: It should! This PR is to unblock https://github.com/square/Valet/pull/174, which bumps to 1.7. CP 1.4 didn’t work on Xcode 10.2 CI: 1.6 was the earliest release that did.

PR #180, opened 2019-06-06
Adopt SPM

dfed:
This PR properly adopts SPM.

In addition to updating our `Package.swift`, I did the following in this PR:
* I separated our Unit tests (tests that test Valet's internals) from our Integration tests (tests that interact with the Keychain) on disk. This allowed me to only run Unit tests in SPM. SPM does not yet support hosted test environments, so running Integration tests with SPM wasn't possible.
* I made `LegacyValet` an internal library and moved it into Sources (and updated the Podspec to not select `LegacyValet`). SPM does not support mixed Objective-C + Swift modules at this time, so I needed to break `LegacyValet` into its own module. The new file structure also conforms to SPM defaults.

Many existing test files were split into two files, one for Unit tests and one for Integration tests. The tests themselves were not changed.


dfed:
ping @gfontenot, @NickEntin and @EricMuller22

dfed:
> Is the change around LegacyValet going to result in different behavior for SPM/Carthage/CocoaPods users?

Good question!

For SPM: Yes. Our prior Package.swift wouldn't build, so the new behavior will be better 🙂
For Carthage: No, because Carthage doesn't build our test code at all
For CocoaPods: No, because our `Valet.podspec` does not include any files from `Sources/LegacyValet` (and it previously did not include code from `Tests/Legacy` either)

PR #181, opened 2019-08-02
Create separate local package for LegacyValet

dfed:
It seems useful to have an example of local dependencies with SPM

cc @xianwen


gfontenot:
Is this the right way to do this? I feel like I'd expect a single `Valet` package file with two targets (`Valet` and `LegacyValet`) being exposed through it. This would also mean we could have `Sources/Valet` and `Sources/LegacyValet` instead of needing to pull `LegacyValet` up to the top level of the repo.

dfed:
@gfontenot since we only want `LegacyValet` to be an implementation detail (it's literally only used for integration testing), making it an un-versioned local Package seemed reasonable. To be entirely honest, I also wanted to use Valet as a test case for playing around with local Package files.

If you think this is a step in the wrong direction, I'm happy to close without merging. There's not much in the way of prior art (that I'm aware of) of Package files for complex repos, so I think we're defining the best practices as we go 😬

gfontenot:
My sense is that the way it exists today (multiple targets, one product) is how SPM intends for us to handle this exact use case. I'm not entirely sure what we actually gain by moving it into a separate package file.

That being said, I don't really have a strong enough opinion here to argue against merging this if you think this is the direction it should go. It _is_ an implementation detail, and so won't (shouldn't) have any actual affect on consumers.

dfed:
> I'm not entirely sure what we actually gain by moving it into a separate package file

What we gain is that we make it more clear that `LegacyValet` is an implementation detail. At least, that's what it looks like to me. I agree this has no direct effect on our customers.

As SPM becomes more actively used in the community + standards develop, we can easily revert this change if it turns out I'm heading in the wrong direction here.

PR #182, opened 2019-08-06
Finish moving files into new LegacyValet folder

dfed:
This PR is a follow-up to https://github.com/square/Valet/pull/181. I had these files moved locally, but I didn't commit. With this PR, the work started in #181 is complete.


dfed:
Here's what the `Package.swift` Xcode project looks like in Xcode 11 beta:
image

Note that the target information is populated (that isn't true on `master`)

PR #186, opened 2019-09-21
Allow Valet's Package.swift to build on all platforms

### LINKED ISSUE #185: LegacyValet build errors using SPM in Xcode 11
When building Valet using SPM in Xcode 11, the project fails to build for watchOS and tvOS. It looks as though `LegacyValet` doesn't build properly on the platforms, and this is specified as a target to build in the `Package.swift`. This is reproducible in the `xcodeproj` by attempting to build the `LegacyValet` project on both watchOS and tvOS.

watchOS:
Screenshot 2019-09-21 at 20 28 23

tvOS:
Screenshot 2019-09-21 at 20 28 32

dfed:
This PR gets Valet's Package.swift to build on all platforms, and adds jobs to CI to ensure we don't break this going forward. Resolves #185.

In this PR I also added a bunch of `xcode11` jobs to CI.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -20,14 +20,7 @@ let package = Package(
         .target(
             name: "Valet",
             dependencies: []),
-        .target(
-            name: "LegacyValet",
-            dependencies: [],
-            publicHeadersPath: "Public"),
-        .testTarget(
-            name: "ValetTests",
-            dependencies: ["Valet"])
     ],
     swiftLanguageVersions: [.v4, .v4_2, .v5]
 )
-let version = Version(3, 2, 5)
+let version = Version(3, 2, 7)
--- a/Sources/Valet/SinglePromptSecureEnclaveValet.swift
+++ b/Sources/Valet/SinglePromptSecureEnclaveValet.swift
@@ -18,6 +18,8 @@
 //  limitations under the License.
 //
 
+#if os(iOS) || os(macOS)
+
 import LocalAuthentication
 import Foundation
 
@@ -312,3 +314,5 @@ extension SinglePromptSecureEnclaveValet {
         }
     }
 }
+
+#endif
--- a/Tests/ValetIntegrationTests/ValetIntegrationTests.swift
+++ b/Tests/ValetIntegrationTests/ValetIntegrationTests.swift
@@ -36,8 +36,15 @@ func testEnvironmentIsSigned() -> Bool {
         
         return false
     }
-    
-    return true
+
+    if let simulatorVersionInfo = ProcessInfo.processInfo.environment["SIMULATOR_VERSION_INFO"],
+        simulatorVersionInfo.contains("iOS 13") || simulatorVersionInfo.contains("tvOS 13")
+    {
+        // Xcode 11's simulator does not support code-signing in a unit-test environment.
+        return false
+    } else {
+        return true
+    }
 }
 
 
--- a/Tests/ValetIntegrationTests/ValetIntegrationTests.swift
+++ b/Tests/ValetIntegrationTests/ValetIntegrationTests.swift
@@ -40,7 +40,7 @@ func testEnvironmentIsSigned() -> Bool {
     if let simulatorVersionInfo = ProcessInfo.processInfo.environment["SIMULATOR_VERSION_INFO"],
         simulatorVersionInfo.contains("iOS 13") || simulatorVersionInfo.contains("tvOS 13")
     {
-        // Xcode 11's simulator does not support code-signing in a unit-test environment.
+        // Xcode 11's simulator does not support code-signing.
         return false
     } else {
         return true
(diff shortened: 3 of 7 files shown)


Package.swift
```diff
@@ -20,14 +20,7 @@ let package = Package(
         .target(
             name: "Valet",
             dependencies: []),
-        .target(
-            name: "LegacyValet",
```
dfed: This target wasn't used by anything in the Package, so it seems reasonable to remove it.

Package.swift
```diff
@@ -20,14 +20,7 @@ let package = Package(
...
-            dependencies: [],
-            publicHeadersPath: "Public"),
-        .testTarget(
-            name: "ValetTests",
```
dfed: Unfortunately, having the `ValetTests` target in the Package caused the `watchOS` build to fail, since `XCTest` does not exist in `watchOS`. The tests are still run via the manually managed `Valet.xcodeproj`, so we don't lose much by removing this test target here.

Package.swift
```diff
@@ -20,14 +20,7 @@ let package = Package(
...
-            dependencies: ["Valet"])
     ],
     swiftLanguageVersions: [.v4, .v4_2, .v5]
 )
-let version = Version(3, 2, 5)
```
dfed: Whoops. We need to remember to keep this version up-to-date.

Note that I'm not bumping the podspec version in this PR, since this PR does not affect the Cocoapods installation.

Sources/Valet/SinglePromptSecureEnclaveValet.swift
```diff
@@ -18,6 +18,8 @@
 //  limitations under the License.
 //
 
+#if os(iOS) || os(macOS)
```
dfed: In our manually managed project we'd previously excluded this file. However, `Package.swift` has no way to exclude a single file in `Sources/Valet` from only a single target. So now we include this file (in both `Package.swift` and `Valet.xcodeproj`), but compile out the contents.

Valet.xcodeproj/project.pbxproj
```diff
@@ -191,6 +191,8 @@
 		16DF6B0D204B496800F8E0A4 /* Valet watchOS Test Host App.app in Embed Watch Content */ = {isa = PBXBuildFile; fileRef = 16DF6AF3204B496800F8E0A4 /* Valet watchOS Test Host App.app */; };
 		16DF6B1F204B4DA600F8E0A4 /* Valet.framework in Resources */ = {isa = PBXBuildFile; fileRef = 16DF6ADA204B45EB00F8E0A4 /* Valet.framework */; };
 		3251BD06224C865E0007453B /* XCTest.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 16D77541215BFC1D004F060C /* XCTest.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
+		32E7115B2336B03800018E15 /* SinglePromptSecureEnclaveValet.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1612FD7922A9CAAB00FC1142 /* SinglePromptSecureEnclaveValet.swift */; };
```
dfed: I added this file to both the `watchOS` and `tvOS` targets in order to make the `Package.swift` and `Valet.xcodeproj` builds more similar.

dfed:
🤔. Looks like Xcode 11 doesn't support code-signing in the unit-test environment. This means all of our tests that expect `testEnvironmentIsSigned()` to be true are failing in Xcode 11.

dfed:
After further investigation: looks like the Xcode 11 simulator isn't properly signed. The ValetTouchIDTest app doesn't work in the simulator, but it does work on device. In the meantime, we have a workaround, and we can still run these tests on Xcode 10 so we're not hosed just yet.

Valet.xcodeproj/project.pbxproj
```diff
@@ -211,20 +183,6 @@
 			remoteGlobalIDString = 1612FDD222A9CB2200FC1142;
 			remoteInfo = "LegacyValet iOS";
 		};
-		1612FDEE22A9CB4800FC1142 /* PBXContainerItemProxy */ = {
-			isa = PBXContainerItemProxy;
-			containerPortal = EA1E1F7B1A8C46080067C991 /* Project object */;
-			proxyType = 1;
-			remoteGlobalIDString = 1612FDE322A9CB3100FC1142;
-			remoteInfo = "LegacyValet watchOS";
```
dfed: `watchOS` support was added in Valet 3. `LegacyValet` is code from Valet 2. This target should never have existed.

Valet.xcodeproj/project.pbxproj
```diff
@@ -211,20 +183,6 @@
...
-		};
-		1612FDFB22A9CC0700FC1142 /* PBXContainerItemProxy */ = {
-			isa = PBXContainerItemProxy;
-			containerPortal = EA1E1F7B1A8C46080067C991 /* Project object */;
-			proxyType = 1;
-			remoteGlobalIDString = 1612FDF522A9CC0700FC1142;
-			remoteInfo = "LegacyValet tvOS";
```
dfed: `tvOS` support was added in Valet 3. `LegacyValet` is code from Valet 2. This target should never have existed.

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -36,8 +36,15 @@ func testEnvironmentIsSigned() -> Bool {
...
     }
-    
-    return true
+
+    if let simulatorVersionInfo = ProcessInfo.processInfo.environment["SIMULATOR_VERSION_INFO"],
+        simulatorVersionInfo.contains("iOS 13") || simulatorVersionInfo.contains("tvOS 13")
+    {
+        // Xcode 11's simulator does not support code-signing.
+        return false
+    } else {
```
dfed: As far as I can tell Xcode 11 doesn't support code-signing in a simulator at all. This means that many of our tests can't be run from Xcode 11.

dfed:
@gfontenot @dnkoutso PTAL when you get a chance.

Package.swift
gfontenot: Bummer

dfed: Agreed. Excited to get these tests back

PR #191, opened 2019-12-22
[Valet 4.0] Adopt Swift 5 syntax, and enable support for SinglePromptSecureEnclaveValet on tvOS

dfed:
Built on top of https://github.com/square/Valet/pull/190. This PR leans into Swift 5 syntax, and utilizes `#canImport` to bring `SinglePromptSecureEnclaveValet` to `tvOS`


Sources/Valet/Accessibility.swift
```diff
@@ -84,18 +84,4 @@ public enum Accessibility: Int, CustomStringConvertible, Equatable {
         
         return accessibilityAttribute as String
     }
-    
-    // MARK: Internal
-    
-    internal static func allValues() -> [Accessibility] {
```
dfed: This was replaced by `CaseIterable`'s `allCases`. The downside of using `CaseIterable` is that the conformance is `public`. However, not having to maintain these values over time is also a plus. Happy to backtrack on this change (and the similar changes below). Thoughts?

Sources/Valet/SecureEnclave.swift
```diff
@@ -21,7 +21,6 @@
 import Foundation
 
 
-@available(macOS 10.11, *)
```
dfed: Our minimum supported target on macOS is macOS 10.11. So there is no need to specify availability in this case.

Sources/Valet/SecureEnclaveAccessControl.swift
```diff
@@ -22,17 +22,17 @@ import Foundation
...
+public enum SecureEnclaveAccessControl: Int, CaseIterable, CustomStringConvertible, Equatable {
     /// Access to keychain elements requires user presence verification via Touch ID, Face ID, or device Passcode. Keychain elements are still accessible by Touch ID even if fingers are added or removed. Touch ID does not have to be available or enrolled.
     /// - version: Available on iOS 8 or later, tvOS 8 or later, watchOS 2.0 or later, and macOS 10.11 or later.
     case userPresence = 1
     
     /// Access to keychain elements requires user presence verification via Face ID, or any finger enrolled in Touch ID. Keychain elements remain accessible via Face ID or Touch ID  after faces or fingers are added or removed. Face ID must be enabled with at least one face enrolled, or Touch ID must be available and at least one finger must be enrolled.
-    /// - version: Available on iOS 9 or later, tvOS 9 or later, watchOS 2.0 or later, and macOS 10.12.1 or later.
+    /// - version: Available on iOS 9 or later, tvOS 11.3 or later, watchOS 2.0 or later, macOS 10.12.1 or later, and macOS application extensions in 10.13.4 or later.
```
dfed: This comment was wrong. Support for this field was added after iOS 9. Here and below.

Sources/Valet/SecureEnclaveAccessControl.swift
```diff
@@ -75,17 +75,17 @@ public enum SecureEnclaveAccessControl: Int, CustomStringConvertible, Equatable
         case .userPresence:
             return .userPresence
         case .biometricAny:
-            if #available(macOS 10.12.1, *) {
-                return .init(rawValue: 2) // .biometryAny with Xcode 9 compatibility.
+            if #available(tvOS 11.3, OSXApplicationExtension 10.13.4, macOS 10.12.1, *) {
```
dfed: When I started using `biometryAny` in the code, I got a build error that I needed a `#available` check for `OSXApplicationExtension 10.13.4`. I added it, and updated the documentation above.

Same comment applies to `biometryCurrentSet` below.

Sources/Valet/SecureEnclaveAccessControl.swift
```diff
@@ -97,9 +97,9 @@ public enum SecureEnclaveAccessControl: Int, CustomStringConvertible, Equatable
             }
         }
     }
-    
-    internal static func allValues() -> [SecureEnclaveAccessControl] {
-        if #available(macOS 10.12.1, *) {
+
+    public static var allCases: Self.AllCases {
```
dfed: We need to manually implement `allCases` since different cases are available depending on the OS.

Sources/Valet/SinglePromptSecureEnclaveValet.swift
```diff
@@ -18,14 +18,14 @@
 //  limitations under the License.
 //
 
-#if os(iOS) || os(macOS)
+#if canImport(LocalAuthentication)
```
dfed: Now that we're no longer supporting Xcode 9, we can use `canImport` to enable `SinglePromptSecureEnclaveValet` on tvOS 10+.

NickEntin: Do we still need the `@available(tvOS 10.0, *)` given the `canImport` here?

Sources/Valet/Valet.swift
```diff
@@ -389,16 +389,15 @@ internal extension Valet {
...
         }
     }
 
     class func iCloudPermutations(with identifier: Identifier, shared: Bool = false) -> [Valet] {
-        return CloudAccessibility.allValues().map { cloudAccessibility in
-            let valet: Valet = shared ? .iCloudSharedAccessGroupValet(with: identifier, accessibility: cloudAccessibility) : .iCloudValet(with: identifier, accessibility: cloudAccessibility)
-            return valet
+        CloudAccessibility.allValues().map { cloudAccessibility in
+            shared ? .iCloudSharedAccessGroupValet(with: identifier, accessibility: cloudAccessibility) : .iCloudValet(with: identifier, accessibility: cloudAccessibility)
+
```
NickEntin: nit: no newline here

Sources/Valet/SinglePromptSecureEnclaveValet.swift
dfed: Yup. If I get rid of it we get build failures 🙂. Here's my understanding as to why:

1.`#if canImport` is a compiler flag that enables a class for compilation in the case where the SDK has access to the library in question.
1. `@available` is a SDK flag that makes a class as available on particular OS versions.

As such, when we're targeting a `tvOS` SDK on Xcode 11, `LocalAuthentication` can always be imported because Xcode 11's SDK has `LocalAuthentication` on `tvOS`. We still need to limit the availability of the class that is now bundled in our `tvOS` library to ensure there are no crashes when running this code on tvOS 9.

PR #192, opened 2019-12-22
[Valet 4.0] Enable running CI on iOS 10, tvOS 10, and watchOS 3

dfed:
This PR enables CI for 3 more platforms that we support.

Note that Xcode 11 installs on Travis do not come with iOS 9, tvOS 9, or watchOS 2. Since our minimum supported Xcode is Xcode 11, we can not run tests against these SDKs in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,15 @@ script:
 matrix:
   include:
   - osx_image: xcode11.3
-    env: ACTION="swift-package";PLATFORMS="iOS_13,tvOS_13,macOS_10_15,watchOS_6";
-  - osx_image: xcode11.3
-    env: ACTION="xcode";PLATFORMS="iOS_12,tvOS_12,watchOS_5";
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -7,13 +7,15 @@ script:
 matrix:
   include:
   - osx_image: xcode11.3
-    env: ACTION="swift-package";PLATFORMS="iOS_13,tvOS_13,macOS_10_15,watchOS_6";
```
dfed: this line moved down with no changes

.travis.yml
```diff
@@ -7,13 +7,15 @@ script:
...
-  - osx_image: xcode11.3
-    env: ACTION="xcode";PLATFORMS="iOS_12,tvOS_12,watchOS_5";
```
dfed: this line moved down with no changes

dfed:
Force push was after rebasing `develop--4.0`

PR #193, opened 2019-12-22
Fix keychains that rely on kSecAttrAccessGroup on macOS Catalina

dfed:
It looks like Catalina broke our logic for finding a `sharedAccessGroupPrefix`. There's new documentation for [kSecAttrAccessGroup](https://developer.apple.com/documentation/security/ksecattraccessgroup) in macOS Catalina:

image

This PR adds [kSecUseDataProtectionKeychain](https://developer.apple.com/documentation/security/ksecusedataprotectionkeychain) to all keychain queries on `macOS`. This follows Apple's recommendation:

image


Sources/Valet/Internal/SecItem.swift
```diff
@@ -70,23 +70,27 @@ internal final class SecItem {
...
+        var query: [CFString : Any] = [
             kSecClass : kSecClassGenericPassword,
             kSecAttrAccount : "SharedAccessGroupAlwaysAccessiblePrefixPlaceholder",
             kSecReturnAttributes : true,
-            kSecAttrAccessible : Accessibility.alwaysThisDeviceOnly.secAccessibilityAttribute
-            ] as CFDictionary
+            kSecAttrAccessible : Accessibility.alwaysThisDeviceOnly.secAccessibilityAttribute,
+        ]
+
+        if #available(macOS 10.15, *) {
```
NickEntin: > The data protection key affects operations only in macOS. Other platforms automatically behave as if the key is set to true, and ignore the key in the query dictionary. You can safely use the key on all platforms.

Probably safe to remove the availability check here.  Fine either way though, since it should do the same thing.

Sources/Valet/Internal/SecItem.swift
```diff
@@ -70,23 +70,27 @@ internal final class SecItem {
...
+        }
         
         secItemLock.lock()
         defer {
             secItemLock.unlock()
         }
         
         var result: AnyObject? = nil
-        var status = SecItemCopyMatching(query, &result)
+        var status = SecItemCopyMatching(query as CFDictionary, &result)
```
NickEntin: I assume this is addressing a warning?

dfed: This is because I got rid of `as CFDictionary` earlier in this method.

Sources/Valet/Internal/SecItem.swift
dfed: I need this check because this field was introduced on macOS 10.15 (we support all the way back to macOS 10.11).

dfed:
> Is this going to affect accessing values set using older versions of Valet

Good catch! I've added a commit that adds this field when setting a value, but doesn't require it when reading values out of the keychain.

Sources/Valet/Internal/Keychain.swift
```diff
@@ -109,6 +109,10 @@ internal final class Keychain {
         var secItemQuery = options
         secItemQuery[kSecAttrAccount as String] = key
         
+        if #available(iOS 13.0, tvOS 13.0, watchOS 6.0, macOS 10.15, *) {
+            secItemQuery["nleg"] = true // kSecUseDataProtectionKeychain for Xcode 9 and Xcode 10 compatibility.
```
NickEntin: Is the `#available` check here not enough to use `kSecUseDataProtectionKeychain`?  If not, should we move this to a `Constants` enum or something along those lines?

Same applies to `SecItem.swift`

dfed: It isn't unfortunately, since this branch still needs to support Xcode 10 and Xcode 9. Neither of those Xcode's know about this field, and the `if #available` check isn't like an `#if` in Objective-C – the code inside still gets compiled if the condition isn't met.

This kinda code sucks, and is one of the things that makes writing Swift code that supports multiple Xcode versions hard. It's one of the reasons I'm really looking forward to the develop--4.0 branch, since I'll be able to use the proper constant here.

I did think about making a local constant for now, but given that I'm going to bring this fix to develop--4.0 with the actual constants as soon as this lands... it didn't seem worth the time 😅.

dfed:
Would love to get this fix into a release so that Mac apps can recover from this issue. Can I get another review?

NickEntin:
Based on testing, it appears that `SecItemAdd` is the only place where the `kSecUseDataProtectionKeychain` property validation is currently occurring. There is some potential for the future appearance of data loss here - specifically if Apple begins to enforce that `kSecUseDataProtectionKeychain` be included when utilizing `SecItemDelete` and `SecItemCopyMatching` as described in the documentation. In this case, items could be written to the Valet, but not read or deleted from it. The fix for this is to specify `kSecUseDataProtectionKeychain` when reading and deleting.

The trade-off here is that if we `kSecUseDataProtectionKeychain` on read, we will cause data loss when data is written to a Valet on macOS 10.14 or earlier and the system is subsequently updated to macOS 10.15. Leaving `kSecUseDataProtectionKeychain` out on read means that we could potentially read in data from a Valet with the same identifier but different accessibility level or access group, since `kSecAttrAccessible` and `kSecAttrAccessGroup` only apply when `kSecUseDataProtectionKeychain` is true.

dfed:
I've updated this PR to always use `kSecUseDataProtectionKeychain`, and I've added `func migrateObjectsFromPreCatalina()` to `Valet`. I need to add a README blurb on how to use that method still.

dfed:
> Leaving `kSecUseDataProtectionKeychain` out on read means that we could potentially read in data from a Valet with the same identifier but different accessibility level or access group, since `kSecAttrAccessible` and `kSecAttrAccessGroup` only apply when `kSecUseDataProtectionKeychain` is true.

Upon further investigation, this isn't entirely true. We encode the accessibility in the `kSecAttrService`, so we can't read data from another Valet with the same identifier but a different accessibility level.

Additionally, since `kSecAttrAccessGroup` is derived _from_ the identifier, it is impossible to have a Valet with the same identifier but different `kSecAttrAccessGroup` value.

That said, I still think protecting against a future where Apple enforces the presence of `kSecUseDataProtectionKeychain` on read.

Sources/Valet/Internal/SecItem.swift
```diff
@@ -70,23 +70,34 @@ internal final class SecItem {
...
             kSecClass : kSecClassGenericPassword,
             kSecAttrAccount : "SharedAccessGroupAlwaysAccessiblePrefixPlaceholder",
             kSecReturnAttributes : true,
-            kSecAttrAccessible : Accessibility.alwaysThisDeviceOnly.secAccessibilityAttribute
-            ] as CFDictionary
-        
+            kSecAttrAccessible : Accessibility.alwaysThisDeviceOnly.secAccessibilityAttribute,
+        ]
+
+        if #available(iOS 13.0, tvOS 13.0, watchOS 6.0, macOS 10.15, *) {
```
dfed: There's no harm in using `kSecUseDataProtectionKeychain` when looking up the `sharedAccessGroupPrefix` on all operating systems.

Sources/Valet/Internal/SecItem.swift
```diff
@@ -70,23 +70,34 @@ internal final class SecItem {
...
+            // Update the account to be distinct, since we're changing the query.
+            query[kSecAttrAccount] = "SharedAccessGroupAlwaysAccessiblePrefixPlaceholderWithDataProtection"
```
dfed: If we don't do this, a `SharedAccessGroupAlwaysAccessiblePrefixPlaceholder` that was added without `kSecUseDataProtectionKeychain` wouldn't be found on `SecItemCopy`, but trying to add that account/data to the keychain would return a `errSecDuplicateItem`.

Sources/Valet/Internal/Service.swift
```diff
@@ -42,9 +42,19 @@ internal enum Service: CustomStringConvertible, Equatable {
     internal func generateBaseQuery() -> [String : AnyHashable] {
         var baseQuery: [String : AnyHashable] = [
             kSecClass as String : kSecClassGenericPassword as String,
-            kSecAttrService as String : secService
+            kSecAttrService as String : secService,
         ]
-        
+
+        #if os(macOS)
+        if #available(macOS 10.15, *) {
```
dfed: I _only_ add `kSecUseDataProtectionKeychain` to our base query on macOS, since macOS is the only OS that _needs_ this value.

Adding this value to the base query is a breaking change, since anything written to the keychain without `kSecUseDataProtectionKeychain` set will not be read when `kSecUseDataProtectionKeychain` is in the base query. I wrote `migrateObjectsFromPreCatalina` to solve this issue on macOS, but decided to allow non-Mac customers to continue using the keychain without migrating objects to include `kSecUseDataProtectionKeychain`.

dfed:
@NickEntin this PR is now ready for review again.

Sources/Valet/Internal/SecItem.swift
dfed: Nevermind. Looks like `errSecDuplicateItem` is only returned when a `kSecAttrService` or `kSecAttrAccount` overlaps. Removing this line.

Sources/Valet/Internal/Service.swift
dfed: After sleeping on it: I realize we can use `kSecUseDataProtectionKeychain` on all platforms. I had [written a test to prove this](https://github.com/square/Valet/pull/193/files#diff-4919d5fe96673b6a2432ee8dc3bbb663R307).

dfed: From [Apple's documentation](https://developer.apple.com/documentation/security/ksecusedataprotectionkeychain):
> It’s highly recommended that you set the value of this key to true for all keychain operations. This key helps to improve the portability of your code across platforms. Use it unless you specifically need access to items previously stored in a legacy keychain in macOS.

Fixed in https://github.com/square/Valet/pull/193/commits/0c9799c491d7b77507be2a0717018ff000b0c0d7

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -284,6 +284,29 @@ class ValetIntegrationTests: XCTestCase
 
         XCTAssertNil(anotherFlavor.string(forKey: "cookie"))
     }
+
+    #if !os(macOS)
+    func test_objectForKey_canReadItemsWithout_kSecUseDataProtectionKeychain_when_kSecUseDataProtectionKeychain_isSetToTrueInKeychainQuery() {
+        let valet = Valet.valet(with: Identifier(nonEmpty: "DataProtectionTest")!, accessibility: .afterFirstUnlock)
+        var dataProtectionWriteQuery = valet.keychainQuery
+        #if swift(>=5.1)
+        preCatalinaWriteQuery[kSecUseDataProtectionKeychain as String] = nil
```
NickEntin: `dataProtectionWriteQuery`?

dfed: 🤦‍♂️ copy/pate is fun
```suggestion
        dataProtectionWriteQuery[kSecUseDataProtectionKeychain as String] = nil
```

PR #194, opened 2019-12-23
Expand test matrix

dfed:
A lot of our unit tests were testing a single `valet` configuration when they could have been testing an entire suite. This PR updates a bunch of our tests to expand the matrix of options we test on any run.

### DIFF
--- a/Sources/Valet/MigrationResult.swift
+++ b/Sources/Valet/MigrationResult.swift
@@ -21,7 +21,7 @@
 import Foundation
 
 @objc(VALMigrationResult)
-public enum MigrationResult: Int, Equatable {
+public enum MigrationResult: Int, Equatable, CustomStringConvertible {
     /// Migration succeeded.
     case success = 1
     /// Migration failed because the keychain query was not valid.
@@ -42,4 +42,21 @@ public enum MigrationResult: Int, Equatable {
     case couldNotWriteToKeychain
     /// Migration failed because removing the migrated data from the keychain failed.
     case removalFailed
+
+    // MARK: CustomStringConvertible
+
+    public var description: String {
+        switch self {
+        case .success: return "success"
+        case .invalidQuery: return "invalidQuery"
+        case .noItemsToMigrateFound: return "noItemsToMigrateFound"
+        case .couldNotReadKeychain: return "couldNotReadKeychain"
+        case .keyInQueryResultInvalid: return "keyInQueryResultInvalid"
+        case .dataInQueryResultInvalid: return "dataInQueryResultInvalid"
+        case .duplicateKeyInQueryResult: return "duplicateKeyInQueryResult"
+        case .keyInQueryResultAlreadyExistsInValet: return "keyInQueryResultAlreadyExistsInValet"
+        case .couldNotWriteToKeychain: return "couldNotWriteToKeychain"
+        case .removalFailed: return "removalFailed"
+        }
+    }
 }
--- a/Tests/ValetIntegrationTests/CloudIntegrationTests.swift
+++ b/Tests/ValetIntegrationTests/CloudIntegrationTests.swift
@@ -26,9 +26,14 @@ import XCTest
 
 class CloudIntegrationTests: XCTestCase
 {
-    static let identifier = Identifier(nonEmpty: "valet_testing")!
+    static let identifier = Valet.sharedAccessGroupIdentifier
     static let accessibility = CloudAccessibility.whenUnlocked
     let valet = Valet.iCloudValet(with: identifier, accessibility: accessibility)
+    var allPermutations: [Valet] {
+        (testEnvironmentIsSigned()
+            ? Valet.iCloudPermutations(with: CloudIntegrationTests.identifier) + Valet.iCloudPermutations(with: ValetIntegrationTests.identifier, shared: true)
+            : [])
+    }
     let key = "key"
     let passcode = "topsecret"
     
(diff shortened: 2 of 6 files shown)


Tests/ValetIntegrationTests/CloudIntegrationTests.swift
```diff
@@ -26,9 +26,14 @@ import XCTest
...
 {
-    static let identifier = Identifier(nonEmpty: "valet_testing")!
+    static let identifier = Valet.sharedAccessGroupIdentifier
     static let accessibility = CloudAccessibility.whenUnlocked
     let valet = Valet.iCloudValet(with: identifier, accessibility: accessibility)
+    var allPermutations: [Valet] {
+        (testEnvironmentIsSigned()
+            ? Valet.iCloudPermutations(with: CloudIntegrationTests.identifier) + Valet.iCloudPermutations(with: ValetIntegrationTests.identifier, shared: true)
+            : [])
+    }
```
dfed: Instead of testing `testEnvironmentIsSigned` below, we test it here.

Tests/ValetIntegrationTests/CloudIntegrationTests.swift
```diff
@@ -67,49 +70,30 @@ class CloudIntegrationTests: XCTestCase
...
-        }
-        
-        Valet.iCloudPermutations(with: valet.identifier).forEach { permutation in
-            XCTAssertTrue(permutation.canAccessKeychain(), "\(permutation) could not access keychain.")
+        allPermutations.forEach { valet in
+            XCTAssertTrue(valet.canAccessKeychain(), "\(valet) could not access keychain.")
         }
     }
-    
-    func test_canAccessKeychain_sharedAccessGroup()
```
dfed: This test was rolled into the previous one

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -563,8 +601,7 @@ class ValetIntegrationTests: XCTestCase
         XCTAssertEqual(noItemsFoundError, anotherFlavor.migrateObjects(matching: valet.keychainQuery, removeOnCompletion: true))
     }
 
-    // FIXME: Looks to me like this test may no longer be valid, need to dig a bit
```
dfed: After some investigation, it seem like iOS 11 and iOS 12 only return synchronizable (iCloud) keychain items if `kSecAttrSynchronizable` is specified in the query (this is unlike iOS 13, and I believe – but I haven't verified – iOS 9 and iOS 10).

therefore, this test would fail, because the value set in `anotherFlavor` wouldn't be returned from the `conflictingQuery`. To fix the test, I've used a Valet with a differing identifier rather than a different flavor.

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -142,33 +144,6 @@ class ValetIntegrationTests: XCTestCase
         }
     }
 
-    // MARK: Equality
```
dfed: These tests already exist in `ValetTests.swift`. I think I mistakenly left them in both files when I split this file in to two for SPM.

Sources/Valet/MigrationResult.swift
```diff
@@ -42,4 +42,21 @@ public enum MigrationResult: Int, Equatable {
     case couldNotWriteToKeychain
     /// Migration failed because removing the migrated data from the keychain failed.
     case removalFailed
+
+    // MARK: CustomStringConvertible
+
+    public var description: String {
```
dfed: adding this helped when debugging tests where the `MigrationResult` return value differed from what was expected

Tests/ValetIntegrationTests/CloudIntegrationTests.swift
```diff
@@ -26,9 +26,14 @@ import XCTest
 
 class CloudIntegrationTests: XCTestCase
 {
-    static let identifier = Identifier(nonEmpty: "valet_testing")!
+    static let identifier = Valet.sharedAccessGroupIdentifier
     static let accessibility = CloudAccessibility.whenUnlocked
     let valet = Valet.iCloudValet(with: identifier, accessibility: accessibility)
```
NickEntin: Why is this `valet` still needed here?

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -73,10 +73,14 @@ internal extension Valet {
 
 class ValetIntegrationTests: XCTestCase
 {
-    static let identifier = Identifier(nonEmpty: "valet_testing")!
+    static let identifier = Valet.sharedAccessGroupIdentifier
     let valet = Valet.valet(with: identifier, accessibility: .whenUnlocked)
```
NickEntin: Is this `valet` still necessary?

Tests/ValetIntegrationTests/CloudIntegrationTests.swift
dfed: Good catch! Updated.

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
dfed: We still need this because we have tests that are exercising the differences between configurations. However, the naming of this property did not make this obvious at all. I've updated for clarity. Good catch.

PR #196, opened 2019-12-23
Do not build branch build on every push on PR branches

dfed:
We don't need to build both a `pr` build and a `push` build on every pull request. This PR makes it so we only build `push` builds for `master` and `develop--4.0`. We'll still have PR builds on every PR, but we won't build every PR build twice.

See [Travis's docs](https://docs.travis-ci.com/user/customizing-the-build/#safelisting-or-blocklisting-branches) for more info on this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,3 +26,8 @@ matrix:
     env: ACTION="pod-lint";SWIFT_VERSION="4.0"
   - osx_image: xcode9
     env: ACTION="carthage"
+
+branches:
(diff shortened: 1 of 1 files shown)


dfed:
Admin merging through checks, since this is just a CI change, and I'd like to get these time savings on my open PRs 🙂. Happy to take feedback post-merge.

NickEntin:
Post-merge LGTM 👍

PR #198, opened 2019-12-25
[Valet 4.0] Convert APIs to throw rather than return booleans or enums

dfed:
This PR tries to make Valet more semantic Swift by relying on `throws` rather than `Bool` or `enum` return types. It's a bit of a massive change, but IMO this API feels a whole lot nicer to use.


Sources/Valet/MigrationError.swift
```diff
@@ -9,27 +9,20 @@
...
 
 @objc(VALMigrationResult)
-public enum MigrationResult: Int, Equatable {
-    /// Migration succeeded.
-    case success = 1
+public enum MigrationError: Int, Error, Equatable {
     /// Migration failed because the keychain query was not valid.
     case invalidQuery
-    /// Migration failed because no items to migrate were found.
-    case noItemsToMigrateFound
```
dfed: Instead we'll throw `KeychainError.itemNotFound`.

Sources/Valet/MigrationError.swift
```diff
@@ -9,27 +9,20 @@
...
-    /// Migration failed because the keychain could not be read.
-    case couldNotReadKeychain
```
dfed: Instead we'll throw `KeychainError.couldNotAccessKeychain`.

Sources/Valet/MigrationError.swift
```diff
@@ -38,8 +31,6 @@ public enum MigrationResult: Int, Equatable {
     case duplicateKeyInQueryResult
     /// Migration failed because a key in the keychain duplicates a key already managed by Valet.
     case keyInQueryResultAlreadyExistsInValet
-    /// Migration failed because writing to the keychain failed.
-    case couldNotWriteToKeychain
```
dfed: This error was never used.

README.md
```diff
@@ -101,17 +101,17 @@ The Accessibility enum is used to determine when your secrets can be accessed. I
...
 ```
 
 ```objc
 NSString *const username = @"Skroob";
-[myValet setString:@"12345" forKey:username];
+[myValet setString:@"12345" forKey:username error:nil];
 NSString *const myLuggageCombination = [myValet stringForKey:username];
 ```
 
-In addition to allowing the storage of strings, Valet allows the storage of `Data` objects via `set(object: Data, forKey key: Key)` and `-objectForKey:`. Valets created with a different class type, via a different initializer, or with a different accessibility attribute will not be able to read or modify values in `myValet`.
```
dfed: welp. This line needed some help. `Key` was part of https://github.com/square/Valet/pull/80, and was removed in https://github.com/square/Valet/pull/80/commits/e54f35bd3b72d0a7bbaa9bf53c5d0e465f3fc274. Also we'd left in the Objective-C description of the method, rather than the Swift one.

Sources/Valet/ErrorHandler.swift
```diff
@@ -1,49 +0,0 @@
...
-//  distributed under the License is distributed on an "AS IS" BASIS,
-//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-//  See the License for the specific language governing permissions and
-//  limitations under the License.
-//
-
-import Foundation
-
-
-public final class ErrorHandler {
```
dfed: We no longer needed a hot-swappable `ErrorHandler`. We now `throw` when an error occurs. We also have `assertionFailure` in a few `internal` methods where the error would very much be a programmer error _within_ Valet.

Sources/Valet/Internal/Keychain.swift
```diff
@@ -48,216 +48,173 @@ internal final class Keychain {
...
         var secItemQuery = options
         secItemQuery[kSecAttrAccount as String] = key
         
-        switch SecItem.deleteItems(matching: secItemQuery) {
-        case .success:
-            return .success
-            
-        case let .error(status):
-            switch status {
-            case errSecInteractionNotAllowed, errSecMissingEntitlement:
```
dfed: This error handling got consolidated in `SecItem.deleteItems`

Sources/Valet/Internal/SecItem.swift
```diff
@@ -21,51 +21,17 @@
...
-    return block()
+    return try block()
 }
 
 
 internal final class SecItem {
     
-    // MARK: Internal Enum
-    
-    internal enum DataResult<SuccessType> {
```
dfed: Excited that we can get rid of these types.

Sources/Valet/Internal/SecItem.swift
```diff
@@ -110,10 +76,9 @@ internal final class SecItem {
     
     // MARK: Internal Class Methods
     
-    internal static func copy<DesiredType>(matching query: [String : AnyHashable]) -> DataResult<DesiredType> {
-        guard query.count > 0 else {
-            ErrorHandler.assertionFailure("Must provide a query with at least one item")
-            return .error(errSecParam)
+    internal static func copy<DesiredType>(matching query: [String : AnyHashable]) throws -> DesiredType {
+        if query.isEmpty {
+            assertionFailure("Must provide a query with at least one item")
```
dfed: We assert, and then continue here. We'll end up throwing a `couldNotAccessKeychain` later in this method. But if we hit this assertion, we have problems _within_ Valet.

Sources/Valet/SinglePromptSecureEnclaveValet.swift
```diff
@@ -303,65 +277,29 @@ public final class SinglePromptSecureEnclaveValet: NSObject {
...
     }
-    
-    // MARK: Public Methods
-    
-    /// - parameter key: A Key used to retrieve the desired object from the keychain.
-    /// - parameter userPrompt: The prompt displayed to the user in Apple's Face ID, Touch ID, or passcode entry UI.
-    /// - returns: The data currently stored in the keychain for the provided key. Returns `nil` if no object exists in the keychain for the specified key, or if the keychain is inaccessible.
-    @available(swift, obsoleted: 1.0)
-    @objc(objectForKey:userPrompt:userCancelled:)
-    public func 🚫swift_object(forKey key: String, withPrompt userPrompt: String, userCancelled: UnsafeMutablePointer<ObjCBool>?) -> Data? {
```
dfed: these methods are no longer necessary, because the `error` parameter will have the information to determine if a user cancelled.

README.md
```diff
@@ -101,17 +101,17 @@ The Accessibility enum is used to determine when your secrets can be accessed. I
 
 ```swift
 let username = "Skroob"
-myValet.set(string: "12345", forKey: username)
+try? myValet.setString("12345", forKey: username)
```
dfed: Thoughts on whether we should show the error being ignored? Both here and in the Objective-C code below. Previously we were discarding the return type, so the current sample-code maintains the prior behavior.

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -421,8 +446,20 @@ class ValetIntegrationTests: XCTestCase
         let removeQueue = DispatchQueue(label: "Remove Object Queue", attributes: .concurrent)
 
         for _ in 1...50 {
-            setQueue.async { XCTAssertTrue(self.valet.set(string: self.passcode, forKey: self.key)) }
-            removeQueue.async { XCTAssertTrue(self.valet.removeObject(forKey: self.key)) }
+            setQueue.async {
+                do {
```
dfed: we need to use a `do/catch` block because we're within an `async` `() -> Void` block that can't `throw`.

dfed:
cc @bachand @fdiaz in case y'all are interested. Trying to modernize the API a bit before Airbnb adopts.

bachand:
Thanks @dfed

Sources/Valet/Internal/SecItem.swift
dfed: (Note we could still get rid of these even if we didn’t move to a `throws` API, since Swift 5 has a built-in `Result` type)

Tests/ValetIntegrationTests/CloudIntegrationTests.swift
```diff
@@ -40,14 +40,10 @@ class CloudIntegrationTests: XCTestCase
     {
         super.setUp()
         
-        ErrorHandler.customAssertBody = { _, _, _, _ in
-            // Nothing to do here.
-        }
-
-        allPermutations.forEach { testValet in testValet.removeAllObjects() }
+        allPermutations.forEach { testValet in try? testValet.removeAllObjects() }
```
NickEntin: Do we care if this fails?  Most importantly, could it cause other tests to fail?

Tests/ValetIntegrationTests/CloudIntegrationTests.swift
```diff
@@ -58,41 +54,47 @@ class CloudIntegrationTests: XCTestCase
...
+            }
         }
     }
     
     // MARK: canAccessKeychain
     
     func test_canAccessKeychain()
     {
-        allPermutations.forEach { valet in
+       allPermutations.forEach { valet in
```
NickEntin: nit: whitespace... this was correct before

Sources/Valet/Internal/Keychain.swift
```diff
@@ -48,216 +48,173 @@ internal final class Keychain {
...
+        let data = try object(forKey: key, options: options)
+        if let string = String(data: data, encoding: .utf8) {
+            return string
+        } else {
+            throw KeychainError.itemNotFound
         }
     }
     
-    internal static func object(forKey key: String, options: [String : AnyHashable]) -> SecItem.DataResult<Data> {
+    internal static func object(forKey key: String, options: [String : AnyHashable]) throws -> Data {
```
NickEntin: What do you think about returning `nil` here when the item is not found, rather than throwing `.itemNotFound`? Feels like throws is more about something going wrong / invalid configuration, where an item not existing yet is a valid case in normal usage.

README.md
NickEntin: Seems reasonable to keep the example simple. The presence of `try?` implies that there is an error case that could (should) be handled in real usage.

Sources/Valet/SecureEnclaveValet.swift
```diff
@@ -122,38 +122,33 @@ public final class SecureEnclaveValet: NSObject {
...
 
     /// - Parameter key: The key to look up in the keychain.
-    /// - Returns: `true` if a value has been set for the given key, `false` otherwise. Will return `false` if the keychain is not accessible.
+    /// - Returns: `true` if a value has been set for the given key, `false` otherwise.
     /// - Note: Will never prompt the user for Face ID, Touch ID, or password.
     @objc
     public func containsObject(forKey key: String) -> Bool {
         execute(in: lock) {
-            guard let keychainQuery = keychainQuery else {
+            guard let keychainQuery = try? keychainQuery() else {
```
NickEntin: Should this method throw if this throws?

Sources/Valet/SecureEnclaveValet.swift
```diff
@@ -271,10 +248,8 @@ public final class SecureEnclaveValet: NSObject {
 extension SecureEnclaveValet {
     
     // MARK: Public Class Methods
-
-    /// - Parameters:
-    ///   - identifier: A non-empty string that uniquely identifies a SecureEnclaveValet.
-    ///   - accessControl: The desired access control for the SecureEnclaveValet.
```
NickEntin: Why'd you remove `accessControl` from the header doc?

dfed: woof good catch. Merge conflict bit me here.

Sources/Valet/SecureEnclaveValet.swift
dfed: I’m not sure! Is `containsObject(forKey` the kind of thing where someone would want/need to understand if an internal error occurred? I guess it’d be good to be able to differentiate between “an error occurred” and “there is no object” though.

Yeah, okay. Let’s make this `throws`. And let’s take the underlying error from `Keychain.containsObject` and throw that too.

(sidebar: I want to get rid of the nullable `keychainQuery` by no longer determining the App ID programmatically. I think enabling App Groups will enable us to remove that & simplify the heck out of this code.)

Sources/Valet/Internal/Keychain.swift
dfed: I like that a lot. I may need to re-create the `noItemsFound` error for the migration case. If I can treat `itemNotFound` consistently everywhere (outside of migration) with nullability, I agree that’s more semantic. Good idea!

Tests/ValetIntegrationTests/CloudIntegrationTests.swift
dfed: fair point that it could cause other tests to fail silently. I’ll make this throw.

Tests/ValetIntegrationTests/CloudIntegrationTests.swift
dfed:
```suggestion
        allPermutations.forEach { valet in
```

Sources/Valet/SecureEnclaveValet.swift
```diff
@@ -28,7 +28,7 @@ public final class SecureEnclaveValet: NSObject {
     // MARK: Public Class Methods
 
     /// - Parameters:
-    ///   - identifier: A non-empty string that uniquely identifies a SecureEnclaveValet.
+    ///   - identifier: A non-empty string that must correspond with the value for keychain-access-groups in your Entitlements file.
```
dfed: This isn't a shared access group Valet:
```suggestion
    ///   - identifier: A non-empty string that uniquely identifies a SecureEnclaveValet.
```

Sources/Valet/SecureEnclaveValet.swift
```diff
@@ -271,10 +248,8 @@ public final class SecureEnclaveValet: NSObject {
 extension SecureEnclaveValet {
     
     // MARK: Public Class Methods
-
-    /// - Parameters:
-    ///   - identifier: A non-empty string that uniquely identifies a SecureEnclaveValet.
-    ///   - accessControl: The desired access control for the SecureEnclaveValet.
+    
+    /// - Parameter identifier: A non-empty string that uniquely identifies a SecureEnclaveValet.
```
dfed:
```suggestion
     /// - Parameters:
     ///   - identifier: A non-empty string that uniquely identifies a SecureEnclaveValet.
     ///   - accessControl: The desired access control for the SecureEnclaveValet.
```

Sources/Valet/SecureEnclaveValet.swift
```diff
@@ -271,7 +248,7 @@ public final class SecureEnclaveValet: NSObject {
 extension SecureEnclaveValet {
     
     // MARK: Public Class Methods
-
+
```
dfed:
```suggestion

```

Sources/Valet/SecureEnclaveValet.swift
dfed: Addressed in 6820428f52986529eca1f9fc42aa49c48847f6f0

dfed: Resolved in ee56d51a0880476a858e8a71b77f4a37f584f3c4

Tests/ValetIntegrationTests/CloudIntegrationTests.swift
dfed: Resolved in ee56d51a0880476a858e8a71b77f4a37f584f3c4

Sources/Valet/Internal/Keychain.swift
dfed: Addressed in 873cda1

dfed: So now that I’ve done it, I’m not convinced I like it. Combining throws and nullability means that accessing a value requires handling two forms of errors. Handling an error in addition to testing for optionality becomes:
```
do {
    guard let object = try valet.objectForKey(“key”) else {
        // handle item not found
        return
    }
    // do thing with object
} catch someErrorWeCareAbout {
    // handle this error
} catch {
    // Nothing to do here, we don’t care about this case
}
```

If we consider “item not found” an error like the other possible errors, the call site code-flow is simpler (no nested indents) since we’re handling all error cases with the same syntax:
```
do {
    let object = try valet.objectForKey(“key”)
    // do thing with object
} catch itemNotFound {
    // handle item not found
} catch someErrorWeCareAbout {
    // handle this error
} catch {
    // Nothing to do here, we don’t care about this case
}
```

Meanwhile, determining “could we find something” reads the same in both the only-error and optional+error approaches:
```
if let object = try? valet.objectForKey(“key”) {
    // do thing with object
}  else {
    // handle item not found
}
```

I’m considering reverting 873cda1. Curious what you think!

NickEntin: Ah interesting. Yeah, after seeing the syntax at the call site I think you're right. In theory, it feels like they should be handled differently; but in practice, I think throwing an error makes for cleaner usage.

dfed: Yeah I'm there with you on the theory; that's what led me to make the change. When reading the tests having to catch just to find `itemNotFound` is also a bit cumbersome. But in practice, I'm guessing folks who care about `itemNotFound` will first try `containsObject(forKey:)`.

The more I think about this, the less convinced I am that `containsObject(forKey:)` is a useful API on a regular `Valet` when `object(forKey:)` and `string(forKey:)` both `throw` specific, catchable errors. Why `try` `containsObject(forKey:)` when you could try accessing the object or string and handle `itemNotFound` in a `catch`? (Note that this API _is_ useful on a `SecureEnclave` or `SinglePromptSecureEnclave` Valet, since accessing the data has a user-prompt side effect and `containsObject` does not.).

Thoughts on whether we should keep `containsObject(forKey:) throws -> Bool` on `Valet`?

In any case, I've reverted 873cda1. @fdiaz curious to know your thoughts on this thread as well.

NickEntin: My gut reaction is to keep `containsObject(forKey:)` on a regular `Valet`. Partly because this keeps the API of the various types of Valets more similar, so you can use the same patterns everywhere.

Totally agreed it doesn't add much when you're simply checking for the presence of a key.
```swift
let hasKey = (try? valet.containsObject(forKey: "key") == true)

let hasKey = (try? valet.object(forKey: "key") != nil)
```

There is still some utility if you're checking for the _absence_ of a key though, since otherwise you need to catch the error to see if it's `.itemNotFound`. To expand on this, say you want to change logic in your method based on the presence of a key, and propagate any errors upwards. With `containsObject(forKey:)` you can call it directly with a `try`, vs. calling `object(forKey:)` in a try/catch and handling one case explicitly.

dfed: Fair point. I guess if you don't want to catch all errors within a method, it's entirely valid to:
```
if try valet.containsObject(forKey: "key") {
    let value = try valet.object(forKey: "key")
    // Do things with value
} else {
    // Do things to prompt the user for a value
}
```

Certainly simpler than:
```
do {
    let value = try valet.object(forKey: "key")
    // Do things with value
} catch KeychainError.itemNotFound {
    // Do things to prompt the user for a value
} catch { error in
    throw error
}
```

Good thinking @NickEntin. Thanks for helping me think through this.

dfed:
I believe Nick's feedback has been fully addressed. This PR should be ready for review again.

Sources/Valet/Internal/Keychain.swift
```diff
@@ -48,216 +48,173 @@ internal final class Keychain {
...
-            }
-        }
+    internal static func removeAllObjects(matching options: [String : AnyHashable]) throws {
+        try SecItem.deleteItems(matching: options)
     }
     
     // MARK: Contains
     
-    internal static func containsObject(forKey key: String, options: [String : AnyHashable]) -> SecItem.Result {
+    internal static func containsObject(forKey key: String, options: [String : AnyHashable]) -> OSStatus {
```
NickEntin: Should we be exposing `OSStatus` in the interface here? Seems like `Keychain` abstracts away the underlying error codes everywhere else except here.

Sources/Valet/Internal/Keychain.swift
```diff
@@ -337,64 +276,63 @@ internal final class Keychain {
...
+                throw MigrationError.dataInQueryResultInvalid
             }
-            
-            guard case let .error(status) = Keychain.containsObject(forKey: key, options: destinationAttributes), status == errSecItemNotFound else {
-                return .keyInQueryResultAlreadyExistsInValet
+
+            if Keychain.containsObject(forKey: key, options: destinationAttributes) == errSecItemNotFound {
+                keysToMigrate.insert(key)
+            } else {
+                throw MigrationError.keyInQueryResultAlreadyExistsInValet
```
NickEntin: Should this throw a different error depending on what `containsObject(...)` returns?

Sources/Valet/Internal/SecItem.swift
```diff
@@ -127,46 +92,36 @@ internal final class SecItem {
...
+            throw KeychainError(status: status)
         }
     }
     
-    internal static func containsObject(matching query: [String : AnyHashable]) -> Result {
+    internal static func containsObject(matching query: [String : AnyHashable]) -> OSStatus {
         guard query.count > 0 else {
-            ErrorHandler.assertionFailure("Must provide a query with at least one item")
-            return .error(errSecParam)
+            // "Must provide a query with at least one item
```
NickEntin: nit: Remove the opening quotation mark

Sources/Valet/Internal/SecItem.swift
```diff
@@ -228,12 +178,20 @@ internal final class SecItem {
...
-            return .error(status)
+            switch KeychainError(status: status) {
+            case .couldNotAccessKeychain:
+                throw KeychainError.couldNotAccessKeychain
+
+            case .missingEntitlement:
+                throw KeychainError.missingEntitlement
+
+            default:
+                // We succeeded as long as we can confirm that the item is not in the keychain.
```
NickEntin: Would all of the other errors mean that this succeeded? Seems like `errSecSuccess` is the only one that means it actually got deleted.

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -181,125 +186,130 @@ class ValetIntegrationTests: XCTestCase
...
 
     // MARK: allKeys
 
-    func test_allKeys()
+    func test_allKeys() throws
     {
-        allPermutations.forEach { valet in
-            XCTAssertEqual(valet.allKeys(), Set(), "\(valet) found keys that should not exist")
+       try allPermutations.forEach { valet in
+            XCTAssertEqual(try vanillaValet.allKeys(), Set(), "\(valet) found keys that should not exist")
```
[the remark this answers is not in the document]
dfed: I can do a mass rename on this file. Seems like we use `vanillaValet` in tests somewhat inconsistently between files.

Sources/Valet/Internal/Keychain.swift
dfed: This is an `internal` method. The call-sites need to inspect the status code in order to make informed decisions. Specifically, the `SecureEnclaveValet` and `SinglePromptSecureEnclaveValet` need to test if an authentication failure error occured, whereas the regular `Valet` does not.

Sources/Valet/Internal/Keychain.swift
dfed: I don't think so? Note that we're currently preserving existing behavior here (we previously returned a `keyInQueryResultAlreadyExistsInValet`, but now we're throwing). If we want to update this logic, let's do that in a separate PR 🙂

Sources/Valet/Internal/SecItem.swift
dfed: This preserves existing behavior that was previously in [Keychain.swift](https://github.com/square/Valet/pull/198/files/b1d205b1c794bae016547e44372f0b1432764d32#diff-15d291e7a67cc811c64966484db0943aL166). Let's consider improving this code in a separate pass to reduce churn in this PR.

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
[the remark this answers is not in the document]
dfed: Resolved in cadeb20

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -181,125 +186,130 @@ class ValetIntegrationTests: XCTestCase
...
     }
     
-    func test_allKeys_doesNotReflectValetImplementationDetails() {
-        allPermutations.forEach { valet in
+    func test_allKeys_doesNotReflectValetImplementationDetails() throws {
+        try allPermutations.forEach { valet in
             // Under the hood, Valet inserts a canary when calling `canAccessKeychain()` - this should not appear in `allKeys()`.
-            _ = valet.canAccessKeychain()
-            XCTAssertEqual(valet.allKeys(), Set())
+            _ = vanillaValet.canAccessKeychain()
```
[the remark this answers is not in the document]
dfed: Resolved in cadeb20

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -181,125 +186,130 @@ class ValetIntegrationTests: XCTestCase
...
 
     // MARK: string(forKey:)
 
-    func test_stringForKey_isNilForInvalidKey()
+    func test_stringForKey_throwsItemNotFoundForKeyWithNoValue() throws
     {
-        allPermutations.forEach { valet in
-            XCTAssertNil(valet.string(forKey: key), "\(valet) found item that should not exit")
+        try allPermutations.forEach { valet in
+            XCTAssertThrowsError(try vanillaValet.string(forKey: key), "\(valet) found item that should not exit") { error in
```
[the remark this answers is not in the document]
dfed: Resolved in cadeb20

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -181,125 +186,130 @@ class ValetIntegrationTests: XCTestCase
...
     }
 
-    func test_stringForKey_retrievesStringForValidKey()
+    func test_stringForKey_retrievesStringForValidKey() throws
     {
-        allPermutations.forEach { valet in
-            XCTAssertTrue(valet.setString(passcode, forKey: key), "\(valet) could not set item in keychain")
-            XCTAssertEqual(passcode, valet.string(forKey: key))
+        try allPermutations.forEach { valet in
+            try vanillaValet.setString(passcode, forKey: key)
```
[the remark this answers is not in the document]
dfed: Resolved in cadeb20

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -313,134 +323,157 @@ class ValetIntegrationTests: XCTestCase
...
         }
     }
     
     // MARK: object(forKey:)
-    
-    func test_objectForKey_isNilForInvalidKey() {
-        allPermutations.forEach { valet in
-            XCTAssertNil(valet.object(forKey: key))
+
+    func test_objectForKey_throwsItemNotFound() throws {
```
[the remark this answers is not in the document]
dfed: Resolved in f78e319

Sources/Valet/SecureEnclaveValet.swift
```diff
@@ -283,10 +257,8 @@ extension SecureEnclaveValet {
         }
         return valet(with: identifier, accessControl: accessControl)
     }
-
-    /// - Parameters:
-    ///   - identifier: A non-empty string that must correspond with the value for keychain-access-groups in your Entitlements file.
-    ///   - accessControl: The desired access control for the SecureEnclaveValet.
```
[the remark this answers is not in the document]
dfed: 🤦‍♂️ thanks for catching this again. Resolved in 7bf227d

Sources/Valet/Internal/Keychain.swift
```diff
@@ -48,216 +48,173 @@ internal final class Keychain {
...
         #else
-            
-            if case .success = containsObject(forKey: key, options: options) {
-                return SecItem.update(attributes: [kSecValueData as String: object], forItemsMatching: secItemQuery)
-            } else {
-                secItemQuery[kSecValueData as String] = object
-                return SecItem.add(attributes: secItemQuery)
-            }
+
+        if containsObject(forKey: key, options: options) == errSecSuccess {
```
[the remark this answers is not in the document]
dfed: Resolved in e60f5d3

Sources/Valet/Internal/SecItem.swift
dfed: Resolved in 2b20666

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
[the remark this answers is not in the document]
dfed: Resolved in cadeb20

Sources/Valet/Internal/Keychain.swift
dfed: Previously, we included the necessary information in `SecItem.Result`. I could theoretically create a new intermediary type (we've deleted `SecItem.Result` in this PR), but since we're not exposing anything publicly that seemed like overkill? I agree that this is a bit of a sharp edge in an otherwise very simple system.

NickEntin: Yeah, that's fair. Given it's all internal, it seems alright.

Sources/Valet/KeychainError.swift
```diff
@@ -0,0 +1,68 @@
...
+public enum KeychainError: Int, CaseIterable, CustomStringConvertible, Error, Equatable {
+    /// The keychain could not be accessed.
+    case couldNotAccessKeychain
+    /// User dismissed the user-presence prompt.
+    case userCancelled
+    /// No data was found for the requested key.
+    case itemNotFound
+    /// The application does not have the proper entitlements to perform the requested action.
+    /// This may be due to an Apple Keychain bug. As a workaround try running on a device that is not attached to a debugger.
+    /// - SeeAlso: https://forums.developer.apple.com/thread/4743"
```
NickEntin: nit: Remove the closing quotation mark

Sources/Valet/SecureEnclave.swift
```diff
@@ -100,71 +55,74 @@ public final class SecureEnclave {
...
         var secItemQuery = options
         secItemQuery[kSecUseAuthenticationUI as String] = kSecUseAuthenticationUIFail
-        
-        switch Keychain.containsObject(forKey: key, options: secItemQuery) {
-        case .success:
+
+        let status = Keychain.containsObject(forKey: key, options: secItemQuery)
+        switch status {
+        case errSecSuccess,
+             errSecInteractionNotAllowed:
```
NickEntin: Might be worth adding a comment explaining why this means the item exists

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -313,134 +323,157 @@ class ValetIntegrationTests: XCTestCase
...
+                XCTAssertEqual(error as? KeychainError, KeychainError.emptyValue)
+            }
         }
-    }
-    
-    func test_setStringForKey_failsForInvalidKey() {
-        allPermutations.forEach { valet in
-            XCTAssertFalse(valet.setString(passcode, forKey: ""))
+
+        func test_setStringForKey_throwsEmptyKeyOnInvalidKey() throws {
```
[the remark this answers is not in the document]
dfed: Woof. Good find! I'm going to blame merge conflicts on this one. Fixing!

Sources/Valet/KeychainError.swift
dfed: Resolved in efeced6

Sources/Valet/SecureEnclave.swift
dfed: Resolved in 883ace0

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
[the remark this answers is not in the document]
dfed: Resolved in 697e53f

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -181,125 +186,130 @@ class ValetIntegrationTests: XCTestCase
...
     }
 
     // MARK: allKeys
 
-    func test_allKeys()
+    func test_allKeys() throws
     {
-        allPermutations.forEach { valet in
-            XCTAssertEqual(valet.allKeys(), Set(), "\(valet) found keys that should not exist")
+       try allPermutations.forEach { valet in
```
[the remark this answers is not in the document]
dfed: Resolved in 697e53f

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -181,125 +186,130 @@ class ValetIntegrationTests: XCTestCase
 
     // MARK: containsObjectForKey
 
-    func test_containsObjectForKey()
+    func test_containsObjectForKey() throws
     {
-        allPermutations.forEach { valet in
-            XCTAssertFalse(valet.containsObject(forKey: key), "\(valet) found object for key that should not exist")
+       try allPermutations.forEach { valet in
```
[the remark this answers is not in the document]
dfed: Resolved in 697e53f

Sources/Valet/Internal/SecItem.swift
```diff
@@ -127,46 +92,36 @@ internal final class SecItem {
...
-            ErrorHandler.assert(status != errSecMissingEntitlement, "A 'Missing Entitlements' error occurred. This is likely due to an Apple Keychain bug. As a workaround try running on a device that is not attached to a debugger.\n\nMore information: https://forums.developer.apple.com/thread/4743")
-            
-            return .error(status)
+            throw KeychainError(status: status)
         }
     }
     
-    internal static func containsObject(matching query: [String : AnyHashable]) -> Result {
+    internal static func containsObject(matching query: [String : AnyHashable]) -> OSStatus {
         guard query.count > 0 else {
```
fdiaz: nit: [Let's use `isEmpty`](https://www.hackingwithswift.com/articles/181/why-using-isempty-is-faster-than-checking-count-0)

Sources/Valet/Internal/Keychain.swift
```diff
@@ -48,216 +48,173 @@ internal final class Keychain {
...
-            }
-        }
+    internal static func removeAllObjects(matching options: [String : AnyHashable]) throws {
+        try SecItem.deleteItems(matching: options)
     }
     
     // MARK: Contains
     
-    internal static func containsObject(forKey key: String, options: [String : AnyHashable]) -> SecItem.Result {
+    internal static func containsObject(forKey key: String, options: [String : AnyHashable]) -> OSStatus {
```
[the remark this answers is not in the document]
fdiaz: Nevermind, I saw below your reply to @NickEntin and also I saw how you're using this and I think it makes sense to keep this not throwing. Maybe we do want to change the method name though since `containsObject` sounds like something that will return a Bool?

Sources/Valet/Internal/SecItem.swift
fdiaz: Let's add a comment here. This does indeed look weird when we know `KeychainError` has more errors that it could handle. Also, should we make this explicit by handling all cases instead of using `default`?

Sources/Valet/SecureEnclaveValet.swift
```diff
@@ -122,127 +122,99 @@ public final class SecureEnclaveValet: NSObject {
...
-    /// - Note: Will never prompt the user for Face ID, Touch ID, or password.
-    @objc
-    public func containsObject(forKey key: String) -> Bool {
-        execute(in: lock) {
-            guard let keychainQuery = keychainQuery else {
-                return false
-            }
-            return SecureEnclave.containsObject(forKey: key, options: keychainQuery)
+    /// - Returns: `true` if a value has been set for the given key, `false` otherwise.
+    /// - Note: Will never prompt the user for Face ID, Touch ID, or password. Method will throw a `KeychainError`.
```
fdiaz: Let's use `Throws` instead of `Note` for the KeychainError comment.

Sources/Valet/SecureEnclaveValet.swift
```diff
@@ -283,7 +257,7 @@ extension SecureEnclaveValet {
         }
         return valet(with: identifier, accessControl: accessControl)
     }
-
+
```
fdiaz: nit: Unnecessary empty space

Tests/ValetTests/MigrationErrorTests.swift
```diff
@@ -0,0 +1,43 @@
...
+//  limitations under the License.
+//
+
+import Foundation
+import XCTest
+
+@testable import Valet
+
+
+class MigrationErrorTests: XCTestCase {
```
fdiaz: nit: `final`

Sources/Valet/MigrationError.swift
```diff
@@ -20,16 +20,11 @@
 
 import Foundation
 
+
```
fdiaz: nit: Empty extra space?

Tests/ValetTests/KeychainErrorTests.swift
```diff
@@ -0,0 +1,64 @@
...
+//  limitations under the License.
+//
+
+import Foundation
+import XCTest
+
+@testable import Valet
+
+
+class KeychainErrorTests: XCTestCase {
```
fdiaz: nit: `final`

Sources/Valet/Internal/Keychain.swift
```diff
@@ -48,216 +48,173 @@ internal final class Keychain {
             var secItemQuery = attributes
             secItemQuery[kSecAttrAccount as String] = canaryKey
             secItemQuery[kSecValueData as String] = Data(canaryValue.utf8)
-            _ = SecItem.add(attributes: secItemQuery)
+            try? SecItem.add(attributes: secItemQuery)
```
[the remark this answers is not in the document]
dfed: the side effect is invisible to the customer. I have a test to prove that the sentinel I'm writing doesn't appear in `allKeys`.

You're right that this is weird, but I think it's okay because the customer never has to know it happened.

Sources/Valet/Internal/Keychain.swift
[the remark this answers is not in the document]
dfed: Fair that the name doesn't match the return type. Thoughts on:
```suggestion
    internal static func performCopy(matchingKey key: String, options: [String : AnyHashable]) -> OSStatus {
```

I'm honestly struggling with this one. What this method does under the hood is perform a `SecItemCopyMatching`, but doesn't retrieve the data. It just needs the `OSStatus` return type from the copy action.

Sources/Valet/Internal/SecItem.swift
dfed: addressed in 249188fdd7f7d62bd337d6067ed0bd051c66d5a6

Sources/Valet/SecureEnclaveValet.swift
dfed: Ah! Missed one! Nice catch 🙂.

Resolved in e7023c9

Sources/Valet/SecureEnclaveValet.swift
dfed: Resolved in d1b2d80

Tests/ValetTests/MigrationErrorTests.swift
dfed:
```suggestion
final class MigrationErrorTests: XCTestCase {
```

Sources/Valet/MigrationError.swift
dfed: Welcome to Square's style guide. Two empty newlines between imports and code 😉

Tests/ValetTests/KeychainErrorTests.swift
dfed:
```suggestion
final class KeychainErrorTests: XCTestCase {
```

Sources/Valet/Internal/SecItem.swift
dfed: I like making it explicit. I think that makes it more clear, and removes the need for a comment. Good thinking!

Addressed in 4e9eb8b

Sources/Valet/Internal/Keychain.swift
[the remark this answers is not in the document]
dfed: Or maybe
```suggestion
    internal static func copyItemStatus(forKey key: String, options: [String : AnyHashable]) -> OSStatus {
```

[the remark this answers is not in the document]
NickEntin: My vote is for `performCopy(matching:)`. The fact is returns a status is implied by the return type.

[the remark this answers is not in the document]
dfed: Addressed in fd5feba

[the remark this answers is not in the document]
fdiaz: Works for me! Thanks for addressing.

Issue #179, opened 2019-06-05
[Public Beta] Valet 4 wishlist


dfed:
I'll be using this issue to track what we'll want to do in our next major version.

- [x] Allow for customizing kSecAttrService on Mac: https://github.com/square/Valet/issues/140
- [x] Remove Xcode 9 Support: #190 
- [x] Dropping Xcode 10 support: #190 
- [x] Drop Swift 4 support: #191
- [ ] ~Consider renaming SecureEnclaveValet per @gravisman – non-biometric keychain items are also stored in the Secure Enclave under the hood. Proposed name: UserPresenceValet~ This item is being removed, since the current naming aligns with [Apple's documentation](https://developer.apple.com/documentation/security/certificate_key_and_trust_services/keys/storing_keys_in_the_secure_enclave).
- [x] Bring `SinglePromptSecureEnclaveValet` to the tvOS target (LocalAuthentication was enabled in tvOS 10): #191
- [x] Get rid of `Accessibility`'s `always` and `alwaysThisDeviceOnly`, since these values have been deprecated: #197
- [x] Move away from `@discardableResult` and towards a `throws` API. We're already letting folk ignore errors. Let's enable that in a more idiomatic way: #198
- [x] Support keychain [sharing via App Groups](https://developer.apple.com/documentation/security/keychain_services/keychain_items/sharing_access_to_keychain_items_among_a_collection_of_apps) in addition to keychain access groups. #230
- [x] Update README to include migration steps required to update to Valet 4: #221

If you want us to include more action items in this list, please file an issue for the change you'd like, and link that issue to this one.

piers12:
The above are all great suggestions but @dfed are you going to still be involved in maintaining Valet?
It seems that you have joined Airbnb now and being the major contributor to Valet, would be great to know how Valet will be maintained going forward?

dfed:
@piers12, I left Square a few years ago at this point, but I absolutely have been and intend to continue staying involved + helping improve Valet. I'm glad the above list is exciting to you! If there's something in particular you'd like from Valet, please [file an issue](https://github.com/square/Valet/blob/master/Contributing.md) 🙂

piers12:
@dfed that's great to hear. I was deciding on wether to use a third party for using keychain in my iOS app and if so which one. The Swifty response from yourself means I think I'll push ahead with using Valet for my own project. Will let you know if there is anything that I would like Valet to add.
Keep up the great work!

dfed:
[...]

PR #199, opened 2019-12-31
[Valet 4.0] Development Branch

dfed:
This branch is the long-lived development branch for Valet 4.0.

https://github.com/square/Valet/issues/179


dfed:
I'm not sure how Codecov is reporting 0% difference between this branch and `master` given that until #208 lands we won't have actual code coverage reporting on `master`.

PR #204, opened 2020-01-16
[Valet 4.0] Documentation cleanup

dfed:
This PR is best reviewed commit-by-commit. Some of this work is also on https://github.com/square/Valet/pull/198 – I'm trying to pull some of the pieces off to make that one easier to review.

I added removing a few `return` statements from the code for good measure 😅

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ NSString *const username = @"Skroob";
 NSString *const myLuggageCombination = [myValet stringForKey:username];
 ```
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -111,7 +111,7 @@ NSString *const username = @"Skroob";
 NSString *const myLuggageCombination = [myValet stringForKey:username];
 ```
 
-In addition to allowing the storage of strings, Valet allows the storage of `Data` objects via `setObject(_ object: Data, forKey key: Key)` and `-objectForKey:`. Valets created with a different class type, via a different initializer, or with a different accessibility attribute will not be able to read or modify values in `myValet`.
```
dfed: There was no reason for this paragraph to discuss Objective-C syntax. It was a holdover from when we switched to Swift.

PR #205, opened 2020-01-16
Fix control list vuln test by using persistent reference

dfed:
I think this test broke in Catalina, ~but I can't be sure –– remember, this test requires a signed environment, so we can't test it in CI~. Update: this test does not require a signed environment and runs on every build.

For some reason, the reference we requested couldn't pull out the vulnerable value from the keychain anymore. I fixed this test by relying on a more-reliable persistent reference.


dfed:
Looks like this did indeed break on Catalina. CI is running on 10.14, and failing. I'll add an `@available` check to use the correct reference type on each platform.

PR #209, opened 2020-01-19
[Valet 4.0] Get SinglePromptSecureEnclaveIntegrationTests running on tvOS

dfed:
This PR is a follow-up to #191. I enabled `SinglePromptSecureEnclave` for the tvOS platform without enabling the tests on that platform.

Note that XCTest doesn't handle mixed availability well – it doesn't respect `@available` flags. So in order to prevent a warning and stop these tests crashing on tvOS 9, I had to turn the `let valet: SinglePromptSecureEnclaveValet` into `func valet() -> SinglePromptSecureEnclaveValet`

I did not include `SinglePromptSecureEnclaveBackwardsCompatibilityTests` in the tvOS target, since tvOS was not supported in Valet 2.0.


dfed:
So, the tests have revealed that `allKeys` doesn't seem to work on a `SinglePromptSecureEnclave` on tvOS 10. The query returns a `-26276`, which isn't anywhere to be found in Apple's header files. According to [secureblackbox.com's list of error codes](https://www.secureblackbox.com/kb/help/ref_err_appleerrorcodes.html), this code represents `errSecInternal`: An internal error occured in the Security framework`.

I have no idea why including an `LAContext` in a query that returns multiple objects would fail only on tvOS 10. My simple solution is to make that API available only on the operating systems we know it works. In https://github.com/square/Valet/pull/209/commits/f6d41b83c97c3bacdaec39c3a9252dbf983e2217 I've made only this API available in tvOS 11+. Does that seem reasonable? Or should we instead make the entire `SinglePromptSecureEnclave` only available on tvOS 11+?

dfed:
Nevermind. It looks like copying anything out of the keychain on tvOS 10 fails with `errSecInternal` when including a `LAContext` in the parameters. I'll make the whole class minimum tvOS 11.

Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
```diff
@@ -20,55 +20,68 @@
...
 
+#if canImport(LocalAuthentication)
 
 class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
 {
     static let identifier = Identifier(nonEmpty: "valet_testing")!
-    let valet = SinglePromptSecureEnclaveValet.valet(with: SinglePromptSecureEnclaveTests.identifier, accessControl: .userPresence)
+
+    @available(tvOS 11.0, *)
+    func valet() -> SinglePromptSecureEnclaveValet {
```
fdiaz: In the PR comment you mentioned that this was done for tvOS 9 support. Is this still needed now that this API is only available on tvOS 11+?

dfed: See this comment for clarification on why I’m making the minimum target tvOS 11: https://github.com/square/Valet/pull/209#issuecomment-576358603

PR #210, opened 2020-01-20
[Valet 4.0] Add explicit tests for CloudAccessibility

dfed:
Can't hurt to explicitly test that CloudAccessibility is mirroring a subset of Accessibility.


Tests/ValetTests/CloudAccessibilityTests.swift
```diff
@@ -0,0 +1,37 @@
...
+final class CloudAccessibilityTests: XCTestCase {
+
+    func test_description_mirrorsAccessibilityCounterpartDescription() {
+        CloudAccessibility.allValues().forEach {
+            XCTAssertEqual($0.description, $0.accessibility.description)
+        }
+    }
+
+    func test_secAccessibilityAttribute_mirrorsAccessibilityCounterpartSecAccessibilityAttribute() {
+        CloudAccessibility.allValues().forEach {
```
fdiaz: Should we make `CloudAccessibility` `CaseIterable`?

dfed: I’ve avoided doing that in the past, since I wanted to limit the possibility that adding a new accessibility type would be a source-breaking change. (Yes, someone COULD switch over the entire accessibility enum, but that seems less likely than enumerating all cases?). 

Note that `allValues` is an `internal` property.

That said, I’m open to `CaseIterable`, since I think I’m convincing myself adding a new case here is already a source-breaking change because of switch statements. Thoughts?

dfed: (We can tackle `CaseIterable` in one fell swoop on all of our enums if we decide to go that way. I still want to know your thoughts!)

fdiaz: I agree that adding a new case will already be a source-breaking change even without the `CaseIterable`. This is something that the [frozen enums](https://github.com/apple/swift-evolution/blob/master/proposals/0192-non-exhaustive-enums.md) change tried to address (but it's not implemented _yey_ for Swift enums), so IMO adding the convenience `CaseIterable` doesn't modify much but gives us the benefit of not having to maintain the allValues manually which is error prone.

dfed: Makes sense. I'll put out a PR once the other outstanding PRs are merged, since doing so now would cause conflicts.

PR #212, opened 2020-01-24
[Valet 4.0] Use CaseIterable instead of allValues where possible

dfed:
This PR follows up on https://github.com/square/Valet/pull/210#discussion_r369661195


Sources/Valet/SecureEnclaveAccessControl.swift
```diff
@@ -24,19 +24,17 @@ import Foundation
 @objc(VALSecureEnclaveAccessControl)
 public enum SecureEnclaveAccessControl: Int, CustomStringConvertible, Equatable {
```
dfed: We couldn't use `CaseIterable` here due to the `@available` tags below. I think using availability here is better than the comments we had before. (We couldn't do this before due to Xcode 9 support)

Issue #213, opened 2020-01-25
Migrating Objects from Pre Catalina unexpected results


gregcotten:
Hi there! I have a macOS 10.15 non-sanboxed, hardened-runtime app _without_ the Keychain Sharing entitlement enabled.

I have a very simple valet setup

```swift
let valet = Valet.valet(with: Identifier(nonEmpty: Bundle.main.bundleIdentifier!)!, accessibility: .whenUnlocked)
```

I have recently upgraded to Valet 3.2.8. Please note this issue below did not occur before the ```kSecUseDataProtectionKeychain = true``` update.

On 10.14 and before, writing to this keychain puts a "application password" in the "login" keychain.

In Catalina, however, writing to this keychain results in a "Missing Entitlement" error. I know this isn't a Valet-specific issue. I go ahead and add the Keychain Sharing entitlement as I read somewhere that that could fix the issue - lo and behold I'm able to write to the keychain! 

However, It's not writing to the "login" keychain but instead the "iCloud" keychain and seems to be using the default Keychain Access Group identifier ($teamID).($appIdentifier) as the access group instead of granting the app access directly. This is all a bit weird, and also ```migrateObjectsFromPreCatalina()``` does not perform an in-place adjustment but instead creates a duplicate keychain item because it now is an "iCloud" item and not a "login" item.

Any insight here is much appreciated!

dfed:
Hi @gregcotten!

Thank you for the report! This is extremely useful information.

The fact that you needed to add a Keychain Sharing entitlement in order to read from that simple Valet setup is not the intended expectation of this library. However, I have reproduced the issue locally. My test application setup has the keychain entitlement (because that's how I test `kSecAttrAccessGroup` keychains) – it didn't occur to me that apps without a keychain sharing entitlement would get a `kSecErrMissingEntitlement` when trying to interact with a simple keychain.

[...]

Issue #140, opened 2018-05-31
Allow kSecAttrService to be a customer-friendly string on Mac


abey:
I'd been working on a project with uses valet and I was curious to know if I could change the name of the location to something more simple.
For example, I would like it to display something like "Storage Space" instead of "VAL_VALValet_initWi...."   

![40080620-5d5bb8f2-5840-11e8-9c6a-3b92af8a1e82](https://user-images.githubusercontent.com/11465329/40800079-34712ad6-64c4-11e8-9b08-67a73841e32b.png)

dfed:
Oh, snap. No, we don't have a way to change the name of the keychain, since that's supposed to be an implementation detail. Honestly, we hadn't thought of this case. I assume the above dialog happens with the `SecureEnclaveValet` (or `SinglePromptSecureEnclaveValet`)?

We can take it as a feature request to allow folks to customize their `kSecAttrService` (where the above string comes from) on these Valets. We'd probably limit said customizability to the Mac code though. Thoughts @EricMuller22?

For now, classifying this as an enhancement request.

gregcotten:
I second this!

gregcotten:
I get this on macOS when I try to set a key/value pair with good old-fashioned `Valet`
```
let valet = Valet.valet(with: Identifier(nonEmpty: Bundle.main.bundleIdentifier!)!, accessibility: .whenUnlocked)
valet.set(string: someString, forKey: someKey)
```

dfed:
@abey & @gregcotten, do you know under what situations macOS prompts the user with the keychain identifier vs not? That info would be really helpful for helping us scope the work we have to do for this issue.

gregcotten:
I will try to do some tests tomorrow!

antons:
As far as I know, this can happen when the code signing identity of the app launched app doesn’t match a know signature, for example when a user tries a Developer ID signed trial, then buys the app from Mac App Store.

dfed:
Thank you for the explanation @antons!

@abey and @gregcotten, does the above jive with your experience?

I’m tempted to say that if the `kSecAttrService` is exposed in the access dialog only if the codesigning is wrong, this issue isn’t worth addressing.

antons:
@dfed I wouldn’t say that the code signing is “wrong”. Developer ID and MAS are both valid ways to sign a public release. The real world example with a trial was meant to underline that I’d like to see the issue fixed as well. It’s not great when implementation details leak into UI like this.

dfed:
[...]

PR #215, opened 2020-01-25
Enable using kSecUseDataProtectionKeychain on macOS keychains prior to Catalina

dfed:
According to my tests in #214, it looks like `kSecUseDataProtectionKeychain` can be used on older macOS installs.

It seems possible that using `kSecUseDataProtectionKeychain` may obviate #140, since `kSecUseDataProtectionKeychain` seems to change how access control works. In the following screenshot, the left item was added to the keychain using `kSecUseDataProtectionKeychain = true`, and the right item was added using `kSecUseDataProtectionKeychain = false`:

image

However, per #213, using `kSecUseDataProtectionKeychain` requires keychain sharing entitlements on macOS, so I've protected the test with `testEnvironmentIsSigned()`. Since our macOS CI environment isn't code-signed, CI can not tell us if this is working. However, given that my `kSecUseDataProtectionKeychain` tests in #214 failed on macOS 10.13 and 10.14 with `errSecMissingEntitlement`, that indicates to me that those keychains understand and respect `kSecUseDataProtectionKeychain` (otherwise we'd have gotten `errSecParam`).

If anyone has a pre-Catalina Mac install around and wants to help me test this, I'd be greatly appreciative. The test would involve using this branch of Valet in your application that has a Keychain Sharing entitlement on a pre-Catalina Mac, running `migrateObjectsFromPreCatalina()` on your Valet, and then trying to access the keychain. I think it'll just work, but, would be best to verify.

### DIFF
--- a/Sources/Valet/Valet.swift
+++ b/Sources/Valet/Valet.swift
@@ -245,8 +245,7 @@ public final class Valet: NSObject, KeychainQueryConvertible {
     #if os(macOS)
     /// Migrates objects that were written to this Valet prior to macOS 10.15 to a format that can be read on macOS 10.15 and later. The new format is backwards compatible, allowing these values to be read on older operating systems.
     /// - returns: Whether the migration succeeded or failed.
-    /// - note: The keychain is not modified if a failure occurs. This method can only be called from macOS 10.15 or later.
-    @available(macOS 10.15, *)
+    /// - note: The keychain is not modified if a failure occurs.
     @objc(migrateObjectsFromPreCatalina)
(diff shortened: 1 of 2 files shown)


dfed:
Actually, on second thought... we probably shouldn't try to set `"nleg"` (i.e. `kSecUseDataProtectionKeychain`) on older macOS keychains, even if it will indeed work. Seems safer to respect Apple's marking of `kSecUseDataProtectionKeychain` as available on 10.15+ and not use the hardcoded constant value on older operating systems. I'm down a rabbit hole. Closing this PR.

PR #216, opened 2020-01-25
Add README documentation explaining that the keychain sharing entitlement is required

### LINKED ISSUE #213: Migrating Objects from Pre Catalina unexpected results

dfed:
This PR should resolve #213.

cc @gregcotten


README.md
```diff
@@ -171,7 +171,9 @@ However, because the Keychain is effectively disk storage, there is no guarantee
...
 
-### Supporting macOS Catalina (10.15) and later
+### Integrating Valet into a macOS application
+
+Your macOS application must have the Keychain Sharing entitlement in order to use Valet, even if your application does not intend to share keychain data between applications. For instructions on how to add a Keychain Sharing entitlement to your application, read [Apple's documentation on the subject](https://developer.apple.com/documentation/security/keychain_services/keychain_items/sharing_access_to_keychain_items_among_a_collection_of_apps). The subsection on "Add Apps to One or More Keychain Access Groups" is particularly helpful, as it includes a screenshot showing Xcode's Signing & Capabilities screen. For more information on why this requirement exists, see [issue #213](https://github.com/square/Valet/issues/213).
```
fdiaz: You can use https://developer.apple.com/documentation/security/keychain_services/keychain_items/sharing_access_to_keychain_items_among_a_collection_of_apps#2974918 to link directly to the "Add Apps to One or More Keychain Access Groups"` section

dfed: That's way better! When you have a chance, I'd love to know how you generated that link. I tried to find link anchors on that page but came up short.

fdiaz: I just inspected the HTML 😁 


image

dfed: Smart!! I'm going to have to steal that trick.

PR #218, opened 2020-02-25
Require that App ID Prefix be explicitly passed into Shared Access Group Valets

dfed:
This PR removes the magic that automatically determined the shared access group.

There are two big drivers behind this PR:
1. #197 made this code significantly more complex, since the app ID prefix could no longer always be accessed. This is discussed some in https://github.com/square/Valet/issues/179#issuecomment-575440030.
1. The auto-magical determining of the App ID prefix is non-deterministic on particular iOS versions when an app has gone through the process of [changing its App ID prefix](https://developer.apple.com/library/archive/technotes/tn2311/_index.html). Airbnb found that on iOS 12, the `sharedAccessGroupPrefix` returned the old App ID, where on iOS 13 the `sharedAccessGroupPrefix` returned our new App ID. This non-determinism was only found on App Store signed builds – debug builds worked. Given the possibility of incorrect behavior by this code, we believe it is best to be explicit.

Note that I need to add this change to the Valet 4 migration guide prior to merging Valet 4 to `master`. I'm tracking that in https://github.com/square/Valet/issues/179#issuecomment-568956038

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,14 +128,14 @@ In addition to allowing the storage of strings, Valet allows the storage of `Dat
 ### Sharing Secrets Among Multiple Applications
 
 ```swift
-let mySharedValet = Valet.sharedAccessGroupValet(with: Identifier(nonEmpty: "Druidia")!, accessibility: .whenUnlocked)
+let mySharedValet = Valet.sharedAccessGroupValet(with: Identifier(nonEmpty: "AppID12345.Druidia")!, accessibility: .whenUnlocked)
 ```
 
 ```objc
-VALValet *const mySharedValet = [VALValet valetWithSharedAccessGroupIdentifier:@"Druidia" accessibility:VALAccessibilityWhenUnlocked];
+VALValet *const mySharedValet = [VALValet valetWithSharedAccessGroupIdentifier:@"AppID12345.Druidia" accessibility:VALAccessibilityWhenUnlocked];
 ```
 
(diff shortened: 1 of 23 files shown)


README.md
```diff
@@ -128,14 +128,14 @@ In addition to allowing the storage of strings, Valet allows the storage of `Dat
 ### Sharing Secrets Among Multiple Applications
 
 ```swift
-let mySharedValet = Valet.sharedAccessGroupValet(with: Identifier(nonEmpty: "Druidia")!, accessibility: .whenUnlocked)
+let mySharedValet = Valet.sharedAccessGroupValet(with: Identifier(nonEmpty: "AppID12345.Druidia")!, accessibility: .whenUnlocked)
```
NickEntin: I think it might be cleaner to create a distinct `SharedAccessGroupIdentifier` that takes two separate strings (the shared access group id and the Valet id), rather than using `Identifier` here. That way it is more obvious what this identifier is, and consumers don't have to know that it needs to be formatted as two strings concatenated with a period.

dfed: I like this a lot. Made code changes in 5078ffc9fdf3847757a0003132153d3f302a3dda. Changes to documentation incoming.

Sources/Valet/Valet.swift
```diff
@@ -377,10 +386,10 @@ public final class Valet: NSObject {
         case .standard:
             serviceAttribute = Service.standard(with: configuration, identifier: identifier, accessibilityDescription: accessibilityDescription)
         #if os(macOS)
-        case .sharedAccessGroupOverride:
-            serviceAttribute = Service.sharedAccessGroup(with: configuration, explicitlySetIdentifier: identifier, accessibilityDescription: accessibilityDescription)
```
dfed: This was a bug before (only on the `develop--4.0` branch).

It occurs to me that I need to enable explicitly set identifiers that use shared access groups more thought. Right now you can only name your explicitly set shared access group to use the same identifier as your keychain entitlement, which isn't terribly human-readable. I can tackle this in a follow-up PR, as this is an existing shortcoming.

PR #219, opened 2020-03-13
Pass on authFailed and userCancelled status.

timothybowers:
Separate authentication failures from user cancelled events.

### DIFF
--- a/Sources/Valet/SecureEnclave.swift
+++ b/Sources/Valet/SecureEnclave.swift
@@ -31,6 +31,8 @@ public final class SecureEnclave {
         case success(Type)
         /// User dismissed the user-presence prompt.
         case userCancelled
+		/// User authentication failed.
+		case authFailed
         /// No data was found for the requested key.
         case itemNotFound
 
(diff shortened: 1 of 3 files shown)


dfed:
Hi @timothybowers! Thank you for your interest in contributing to Valet 🙂

Unfortunately, this issue is one that has been discussed at length before. Searching for [authFailed in this repo](https://github.com/square/Valet/issues?q=authFailed) should reveal everything, but I recommend looking at https://github.com/square/Valet/pull/152, https://github.com/square/Valet/issues/143, and https://github.com/square/Valet/pull/73#issuecomment-184892249. 

It's also worth reading our [contributing guidelines](https://github.com/square/Valet/blob/master/Contributing.md) to familiarize yourself with how to set yourself up for success in this repo.

Given the resources linked to above, this isn't a PR that we'd accept. I'm going to close.

That said, I'd love to hear what your use-case is for separating out `userCancelled` and `authFailed` is, in case we can provide some guidance.

timothybowers:
Hi @dfed!  Thank you for the feedback, and thank you for a great project! 💙

I have looked into this matter again and I agree with you 100% that we do not need authFailed.

I am unable to get the device to generate me an "errSecAuthFailed" error no matter how many times I input the incorrect pin, touchID or faceID.
The only way I can get that error is when I press the power button on the device when it's waiting for authentication. Which to me is the same as userCancelled.

When the user inputs the wrong pin/touchId/faceId, the device merely keeps on asking for you to try again.

Thank you!

Issue #226, opened 2020-05-18
Support arbitrary transforms of data during migration


sergstav:
Thank you for this great library.

I try to do simple migration my old keychain data to Valet, but it gives me an error: keyInQueryResultInvalid

This is my source code:
```
 func migrateOldKeychainDataToValet(key: String) {
        
        let query = [
        kSecClass as String       : kSecClassGenericPassword,
        kSecAttrAccount as String : key,
        kSecAttrService as String : Bundle.main.bundleIdentifier ?? "SwiftKeychainWrapper",
        kSecMatchLimit as String  : kSecMatchLimitAll ] as [String : AnyHashable]
        
        let result = KeychainWrapper.standart.migrateObjects(matching: query, removeOnCompletion: false)
        print("result: \(result)")
    }
```

Where 
`KeychainWrapper.standart.migrateObjects(matching: query, removeOnCompletion: false) `
is just a wrapper of Valet's method.
```
public func migrateObjects(matching query: [String : AnyHashable], removeOnCompletion: Bool) -> MigrationResult {
        return valet.migrateObjects(matching: query, removeOnCompletion: removeOnCompletion)
    }
```
As I understand issue appear on this line
![Screenshot 2020-05-18 at 21 17 07](https://user-images.githubusercontent.com/3393265/82246237-f712f380-994c-11ea-9915-6524d2acf29b.png)

So this line returns error.
```
guard !key.isEmpty else {
                return .keyInQueryResultInvalid
            }
```

I'm doing something wrong?

dfed:
Thank you for the report @sergstav! Love all the context you gave. Can you:
1. Let us know what `key` you are passing into the `key` parameter in the `migrateOldKeychainDataToValet(key:)` method.
1. Print the `keychainEntry` that is causing this issue? Feel free to redact the values in the dictionary, but please do leave the type information so I can understand what this value is.
1. Optional: let us know how you stored data in the keychain prior to Valet.

I'm also curious why you're running migrations on a per-key basis. I'd expect that you could write a method without `kSecAttrAccount as String : key,` in your migration query to migrate all key:value pairs that belong to your `kSecAttrService`. But that said, if you already have an entry in the keychain with an empty `kSecAttrAccount`, you may have to delete that first. Assuming that you don't need the keychain entry that has no associated account, you could consider writing:
```
func migrateOldKeychainDataToValet() {
    let query = [
        kSecClass as String       : kSecClassGenericPassword,
[...]

PR #227, opened 2020-05-20
Rename MigrationError cases with `InQueryResult` to `ToMigrate`

dfed:
This PR updates the case names in `MigrationError` to be more future-proofed. Previously, we explicitly specified that something was wrong with the query result to be migrated. Now we're committing to something begin wrong in the key/value pairs staged to be migrated.

This PR helps us achieve #226 by making the names of these cases less specific to the current migration API. Since this PR represents a breaking API change, it'd be best to get it in before we release Valet 4.

### DIFF
--- a/Sources/Valet/MigrationError.swift
+++ b/Sources/Valet/MigrationError.swift
@@ -25,13 +25,13 @@ import Foundation
 public enum MigrationError: Int, CaseIterable, CustomStringConvertible, Error, Equatable {
     /// Migration failed because the keychain query was not valid.
     case invalidQuery
-    /// Migration failed because a key in the query result could not be read.
+    /// Migration failed because a key staged for migration was empty.
     case keyToMigrateInvalid
-    /// Migration failed because some data in the query result could not be read.
+    /// Migration failed because some data staged for migration was empty.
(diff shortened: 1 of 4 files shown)


Sources/Valet/MigrationError.swift
```diff
@@ -25,14 +25,14 @@ import Foundation
...
     case invalidQuery
-    /// Migration failed because a key in the query result could not be read.
-    case keyInQueryResultInvalid
-    /// Migration failed because some data in the query result could not be read.
-    case dataInQueryResultInvalid
-    /// Migration failed because two keys with the same value were found in the keychain.
-    case duplicateKeyInQueryResult
-    /// Migration failed because a key in the keychain duplicates a key already managed by Valet.
-    case keyInQueryResultAlreadyExistsInValet
+    /// Migration failed because a key staged for migration was empty.
```
NickEntin, line 28: I think we should leave this comment a bit more vague, since it can be invalid for reasons other than being empty (e.g. the key is a type other than a String, so we can’t parse it).

dfed: That's entirely fair. Updating.

PR #228, opened 2020-05-20
Add ability to transform values on migrate

### LINKED ISSUE #226: Support arbitrary transforms of data during migration

dfed:
This PR resolves #226 by adding the ability to `compactMap` over key:value pairs as they are being migrated.

I do not intend to merge this PR prior to the release of Valet 4.0.

### DIFF
--- a/Sources/Valet/Internal/Keychain.swift
+++ b/Sources/Valet/Internal/Keychain.swift
@@ -269,7 +269,11 @@ internal final class Keychain {
         // Sanity check that we are capable of migrating the data.
         var keyValuePairsToMigrate = [String: Data]()
         for keychainEntry in retrievedItemsToMigrateWithData {
-            guard let key = keychainEntry[kSecAttrAccount as String], key as? String != Keychain.canaryKey else {
+            guard let key = keychainEntry[kSecAttrAccount as String] else {
+                throw MigrationError.keyToMigrateInvalid
+            }
+
+            guard key as? String != Keychain.canaryKey else {
                 // We don't care about this key. Move along.
                 continue
             }
--- a/Sources/Valet/Internal/Keychain.swift
+++ b/Sources/Valet/Internal/Keychain.swift
@@ -307,29 +307,25 @@ internal final class Keychain {
             }
         }
 
-        // All looks good. Time to actually migrate.
-        var alreadyMigratedKeys = [String]()
-        func revertMigration() {
-            // Something has gone wrong. Remove all migrated items.
-            for alreadyMigratedKey in alreadyMigratedKeys {
-                try? Keychain.removeObject(forKey: alreadyMigratedKey, options: destinationAttributes)
-            }
-        }
+        // Capture the keys in the destination prior to migration beginning.
+        let keysInKeychainPreMigration = Set(try Keychain.allKeys(options: destinationAttributes))
 
+        // All looks good. Time to actually migrate.
         for keyValuePair in keyValuePairsToMigrate {
             do {
                 try Keychain.setObject(keyValuePair.value, forKey: keyValuePair.key, options: destinationAttributes)
-                alreadyMigratedKeys.append(keyValuePair.key)
             } catch {
-                revertMigration()
+                revertMigration(into: destinationAttributes, keysInKeychainPreMigration: keysInKeychainPreMigration)
                 throw error
             }
         }
     }
 
     internal static func migrateObjects(matching query: [String : AnyHashable], into destinationAttributes: [String : AnyHashable], removeOnCompletion: Bool) throws {
+        // Capture the keys in the destination prior to migration beginning.
         let keysInKeychainPreMigration = Set(try Keychain.allKeys(options: destinationAttributes))
 
+        // Attempt migration.
         try migrateObjects(matching: query, into: destinationAttributes) { keychainKeyValuePair in
             guard let key = keychainKeyValuePair.key as? String else {
                 throw MigrationError.keyToMigrateInvalid
@@ -342,17 +338,21 @@ internal final class Keychain {
             do {
                 try Keychain.removeAllObjects(matching: query)
             } catch {
-                if let allKeysPostPotentiallyPartialMigration = try? Keychain.allKeys(options: destinationAttributes) {
-                    let migratedKeys = allKeysPostPotentiallyPartialMigration.filter { !keysInKeychainPreMigration.contains($0) }
-                    migratedKeys.forEach { migratedKey in
-                        try? Keychain.removeObject(forKey: migratedKey, options: destinationAttributes)
-                    }
-                }
+                revertMigration(into: destinationAttributes, keysInKeychainPreMigration: keysInKeychainPreMigration)
 
                 throw MigrationError.removalFailed
             }
 
             // We're done!
+        }
+    }
+
+    internal static func revertMigration(into destinationAttributes: [String : AnyHashable], keysInKeychainPreMigration: Set<String>) {
+        if let allKeysPostPotentiallyPartialMigration = try? Keychain.allKeys(options: destinationAttributes) {
+            let migratedKeys = allKeysPostPotentiallyPartialMigration.filter { !keysInKeychainPreMigration.contains($0) }
+            migratedKeys.forEach { migratedKey in
+                try? Keychain.removeObject(forKey: migratedKey, options: destinationAttributes)
+            }
         }
     }
 }
--- a/Sources/Valet/Internal/Keychain.swift
+++ b/Sources/Valet/Internal/Keychain.swift
@@ -174,55 +174,55 @@ internal final class Keychain {
     }
     
     // MARK: Migration
-    
-    internal static func migrateObjects(matching query: [String : AnyHashable], into destinationAttributes: [String : AnyHashable], removeOnCompletion: Bool) throws {
+
+    internal static func migrateObjects(matching query: [String : AnyHashable], into destinationAttributes: [String : AnyHashable], compactMap: (MigratableKeyValuePair<AnyHashable>) throws -> MigratableKeyValuePair<String>?) throws {
         guard !query.isEmpty else {
             // Migration requires secItemQuery to contain values.
             throw MigrationError.invalidQuery
         }
-        
+
         guard query[kSecMatchLimit as String] as? String as CFString? != kSecMatchLimitOne else {
             // Migration requires kSecMatchLimit to be set to kSecMatchLimitAll.
             throw MigrationError.invalidQuery
         }
-        
+
         guard query[kSecReturnData as String] as? Bool != true else {
             // kSecReturnData is not supported in a migration query.
             throw MigrationError.invalidQuery
         }
-        
+
         guard query[kSecReturnAttributes as String] as? Bool != false else {
             // Migration requires kSecReturnAttributes to be set to kCFBooleanTrue.
             throw MigrationError.invalidQuery
         }
-        
+
         guard query[kSecReturnRef as String] as? Bool != true else {
             // kSecReturnRef is not supported in a migration query.
             throw MigrationError.invalidQuery
         }
-        
+
         guard query[kSecReturnPersistentRef as String] as? Bool != false else {
             // Migration requires kSecReturnPersistentRef to be set to kCFBooleanTrue.
             throw MigrationError.invalidQuery
         }
-        
+
         guard query[kSecClass as String] as? String as CFString? == kSecClassGenericPassword else {
             // Migration requires kSecClass to be set to kSecClassGenericPassword to avoid data loss.
             throw MigrationError.invalidQuery
         }
-        
+
         guard query[kSecAttrAccessControl as String] == nil else {
             // kSecAttrAccessControl is not supported in a migration query. Keychain items can not be migrated en masse from the Secure Enclave.
             throw MigrationError.invalidQuery
         }
-        
+
         var secItemQuery = query
         secItemQuery[kSecMatchLimit as String] = kSecMatchLimitAll
         secItemQuery[kSecReturnAttributes as String] = true
         secItemQuery[kSecReturnData as String] = false
         secItemQuery[kSecReturnRef as String] = false
         secItemQuery[kSecReturnPersistentRef as String] = true
-        
+
         let collection: Any = try SecItem.copy(matching: secItemQuery)
         let retrievedItemsToMigrate: [[String: AnyHashable]]
         if let singleMatch = collection as? [String : AnyHashable] {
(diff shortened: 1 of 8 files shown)


Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -708,7 +708,7 @@ class ValetIntegrationTests: XCTestCase
         }
     }
 
-    func test_migrateObjectsMatching_withAccountNameAsData_doesNotRaiseException() throws
```
dfed, line 711: This test name was incorrect.

Sources/Valet/Internal/Keychain.swift
```diff
@@ -265,73 +265,94 @@ internal final class Keychain {
                 throw error
             }
         }
-        
+
         // Sanity check that we are capable of migrating the data.
-        var keysToMigrate = Set<String>()
+        var keyValuePairsToMigrate = [String: Data]()
         for keychainEntry in retrievedItemsToMigrateWithData {
-            guard let key = keychainEntry[kSecAttrAccount as String] as? String, key != Keychain.canaryKey else {
```
dfed, line 272: We'll force the `key` to be a `String` later on as part of the `compactMap`. For now, we just want to make sure we _have_ a key.

Tests/ValetObjectiveCBridgeTests/VALValetTests.m
```diff
@@ -167,6 +172,271 @@ - (void)test_iCloudValetWithSharedAccessGroupIdentifier_accessibility_returnsNil
     XCTAssertNil(valet);
 }
 
+- (void)test_containsObjectForKey_returnsTrueWhenObjectExistsInKeychain;
```
dfed, line 175: I realized while I was here that I didn’t have an explicit objective-c test for this wrapper method, so I wrote a couple

Tests/ValetObjectiveCBridgeTests/VALValetTests.m
```diff
@@ -45,6 +45,11 @@ - (NSString *)sharedAccessGroupIdentifier;
 #endif
 }
 
+- (BOOL)testEnvironmentIsSigned;
+{
+    return NSBundle.mainBundle.bundleIdentifier != nil && ![NSBundle.mainBundle.bundleIdentifier isEqualToString:@"com.apple.dt.xctest.tool"];
```
dfed, line 50: This logic is the same logic used in the swift tests

Tests/ValetIntegrationTests/KeychainIntegrationTests.swift
```diff
@@ -0,0 +1,60 @@
+//
+//  KeychainIntegrationTests.swift
+//  Valet iOS
+//
+//  Created by Dan Federman on 5/20/20.
+//  Copyright © 2010 Square, Inc.
```
NickEntin, line 6:
```suggestion
//  Copyright © 2020 Square, Inc.
```

Tests/ValetIntegrationTests/KeychainIntegrationTests.swift
```diff
@@ -0,0 +1,60 @@
...
+//  limitations under the License.
+//
+
+import Foundation
+import XCTest
+
+@testable import Valet
+
+
+class KeychainIntegrationTests: XCTestCase {
```
NickEntin, line 27: nit: `final`
```suggestion
final class KeychainIntegrationTests: XCTestCase {
```

Tests/ValetIntegrationTests/KeychainIntegrationTests.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    func test_revertMigration_removesAllMigratedKeys() throws {
+        guard testEnvironmentIsSigned() else {
+            return
```
NickEntin, line 31: Should we `XCTFail` inside here to make it clear that the test didn't actually run and pass?

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -732,7 +732,130 @@ class ValetIntegrationTests: XCTestCase
...
+                return nil
+            }
+            if key == key1 {
+                return nil
+            } else {
+                return MigratableKeyValuePair<String>(key: key, value: input.value)
+            }
+        }
+
+        XCTAssertNil(try? anotherValet.string(forKey: key1))
```
NickEntin, line 789:
```suggestion
        XCTAssertFalse(try anotherValet.containsObject(forKey: key1))
```

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -732,7 +732,130 @@ class ValetIntegrationTests: XCTestCase
...
+            if key == key1 {
+                return MigratableKeyValuePair<String>(key: key, value: input.value)
+            } else {
+                struct FakeError: Error {}
+                throw FakeError()
+            }
+        }
+
+        XCTAssertNil(try? anotherValet.string(forKey: key1))
+        XCTAssertNil(try? anotherValet.string(forKey: key2))
```
NickEntin, line 825:
```suggestion
        XCTAssertFalse(try anotherValet.containsObject(forKey: key1))
        XCTAssertFalse(try anotherValet.containsObject(forKey: key2))
```

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -732,7 +732,130 @@ class ValetIntegrationTests: XCTestCase
             XCTAssertEqual(error as? MigrationError, .keyToMigrateInvalid)
         }
     }
-    
+
+    func test_migrateObjectsMatchingCompactMap_successfullyMigratesTransformedValue() throws {
```
NickEntin, line 736: Should we also add a test that transforms the `key`? (Same for the Valet -> Valet equivalent below)

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -863,4 +986,115 @@ class ValetIntegrationTests: XCTestCase
...
+                return nil
+            }
+            if key == key1 {
+                return nil
+            } else {
+                return MigratableKeyValuePair<String>(key: key, value: input.value)
+            }
+        }
+
+        XCTAssertNil(try? anotherValet.string(forKey: key1))
```
NickEntin, line 1036:
```suggestion
        XCTAssertFalse(try anotherValet.containsObject(forKey: key1))
```

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -863,4 +986,115 @@ class ValetIntegrationTests: XCTestCase
...
+            if key == key1 {
+                return MigratableKeyValuePair<String>(key: key, value: input.value)
+            } else {
+                struct FakeError: Error {}
+                throw FakeError()
+            }
+        }
+
+        XCTAssertNil(try? anotherValet.string(forKey: key1))
+        XCTAssertNil(try? anotherValet.string(forKey: key2))
```
NickEntin, line 1069:
```suggestion
        XCTAssertFalse(try anotherValet.containsObject(forKey: key1))
        XCTAssertFalse(try anotherValet.containsObject(forKey: key2))
```

Sources/Valet/Valet.swift
```diff
@@ -603,6 +629,53 @@ extension Valet {
         return containsObject
     }
 
+    /// Migrates objects matching the input query into the receiving Valet instance.
+    /// - Parameters:
+    ///   - query: The query with which to retrieve existing keychain data via a call to SecItemCopyMatching.
+    ///   - compactMap: A closure that transforms a key:value pair from the raw pair currently in the keychain into a key:value pair we'll insert into the destination Valet. Returning `nil` from this closure will cause that key:value pair not to be migrated.
```
NickEntin, line 635: Should this mention the `preventMigration` functionality?

Sources/Valet/Valet.swift
```diff
@@ -603,6 +629,53 @@ extension Valet {
...
+    /// - Throws: An error of type `KeychainError` or `MigrationError`.
```
NickEntin, line 636: Does this get converted over to the `error` parameter description automatically?

Sources/Valet/Valet.swift
dfed:
```suggestion
    ///   - compactMap: A closure that transforms a key:value pair from the raw pair currently in the keychain into a key:value pair we'll insert into the destination Valet. Returning `nil` from this closure will cause that key:value pair not to be migrated. Returning `VALMigratableKeyValuePairOutput.preventMigration` will prevent migrating any key:value pairs.
```

Sources/Valet/Valet.swift
dfed: It does not. Good call out!
```suggestion
    ///   - error: An error of type `KeychainError` or `MigrationError`.
```

Sources/Valet/Valet.swift
```diff
@@ -603,6 +629,53 @@ extension Valet {
...
+    @objc(migrateObjectsMatching:compactMap:error:)
+    public func 🚫swift_migrateObjects(matching query: [String : AnyHashable], compactMap: (ObjectiveCCompatibilityMigratableKeyValuePairInput) -> ObjectiveCCompatibilityMigratableKeyValuePairOutput?) throws {
+        try objc_compatibility_migrateObjects(matching: query, compactMap: compactMap)
+    }
+
+    /// Migrates objects matching the input query into the receiving Valet instance.
+    /// - Parameters:
+    ///   - valet: The Valet used to retrieve the existing keychain data that should be migrated.
+    ///   - compactMap: A closure that transforms a key:value pair from the raw pair currently in the keychain into a key:value pair we'll insert into the destination Valet. Returning `nil` from this closure will cause that key:value pair not to be migrated.
+    /// - Throws: An error of type `KeychainError` or `MigrationError`. Will rethrow any error thrown by `compactMap`.
```
dfed, line 648:
```suggestion
    ///   - error: An error of type `KeychainError` or `MigrationError`.
```

Tests/ValetIntegrationTests/KeychainIntegrationTests.swift
dfed: This test will always run. We'll hit this `return` when we run Mac tests in CI. That is expected.

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
dfed: Good call! Will add

dfed: e990014669e0e8fa072aaef73403d0095497c6b7

Sources/Valet/MigratableKeyValuePair.swift
```diff
@@ -0,0 +1,133 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+/// A struct that represented a key:value pair that can be migrated.
+public struct MigratableKeyValuePair<KeyType: Hashable>: Hashable {
```
fdiaz, line 24: Maybe?

```suggestion
public struct MigratableKeyValuePair<Key: Hashable>: Hashable {
```

https://swift.org/documentation/api-design-guidelines#name-according-to-roles

Sources/Valet/MigratableKeyValuePair.swift
```diff
@@ -0,0 +1,133 @@
...
+        self.key = key
+        self.value = Data(stringValue.utf8)
+        preventMigration = false
+    }
+
+    // MARK: Public Static Methods
+
+    /// A sentinal `ObjectiveCCompatibilityMigratableKeyValuePairOutput` that conveys that the migration should be prevented.
+    @objc
+    public static func preventMigration() -> ObjectiveCCompatibilityMigratableKeyValuePairOutput {
```
fdiaz, line 107: This method sounds to me more than its actually doing the action to prevent migration on an instance rather than creating an instance that prevents the migration.

Maybe `migrationDisabled` / `migrationPrevented`?

dfed: Yeah the tough thing here is it isn't super clear that we're doing this work _before_ the migration has started. This is why I wanted to use a future tense. A past-tense phrase here might make it read as if returning it prevented/disabled the migration of just this pair (or future pairs).

In the context where it's used, I hope this method reads well:
https://github.com/square/Valet/blob/c7cec7516d89d82768f5cff14b2d599d54304871/Tests/ValetObjectiveCBridgeTests/VALValetTests.m#L475-L484

The documentation on the `migrateObjectsMatching` Objective-C method also makes it clear how to use this method:
https://github.com/square/Valet/blob/c7cec7516d89d82768f5cff14b2d599d54304871/Sources/Valet/Valet.swift#L661-L668

While I agree that reading this method by itself is a bit confusing, I think within the context it's supposed to be used it makes sense? Let me know if not @fdiaz.

dfed: That said, I should make this method available only to Objective-C

dfed:
Force push had no functional changes. Only change was branch history after merging `develop--4.0` into `master`

PR #230, opened 2020-05-20
Support sharing keychain items using App Groups

### LINKED ISSUE #222: Support sharing keychain items using App Groups
Currently, we only support sharing keychain items with Keychain Access Groups.

We should also support keychain sharing via App Groups.

More information on the differences between Keychain Access Groups and App Groups can be found in Apple's [documentation](https://developer.apple.com/documentation/security/keychain_services/keychain_items/sharing_access_to_keychain_items_among_a_collection_of_apps).

dfed:
This PR resolves #222. This PR is best reviewed commit-by-commit.


Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
```diff
@@ -160,6 +160,8 @@ class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
         }
     }
 
+    #if !os(macOS)
```
NickEntin, line 163: Can we use an `@available` here instead?

Sources/Valet/SecureEnclaveValet.swift
```diff
@@ -262,6 +262,20 @@ extension SecureEnclaveValet {
         return sharedGroupValet(with: identifier, accessControl: accessControl)
     }
 
+    /// - Parameters:
+    ///   - groupPrefix: On iOS, iPadOS, watchOS, and tvOS, this prefix must equal "group". On macOS, this prefix is the application's App ID prefix, which  can be found by inspecting the application's provisioning profile, or viewing the application's App ID Configuration on developer.apple.com. This string must not be empty.
+    ///   - identifier: An identifier that cooresponds to a value in com.apple.security.application-groups in the application's Entitlements file. This string must not be empty.
```
NickEntin, line 267:
```suggestion
    ///   - identifier: An identifier that corresponds to a value in com.apple.security.application-groups in the application's Entitlements file. This string must not be empty.
```

Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
dfed: We can't use `@available` or `#available` to exclude a platform. Both require a trailing `*` which opts into all platforms that are not mentioned.

Valet.xcodeproj/project.pbxproj
```diff
@@ -2516,10 +2520,11 @@
 				CLANG_WARN_INFINITE_RECURSION = YES;
 				CLANG_WARN_SUSPICIOUS_MOVE = YES;
 				CODE_SIGN_ENTITLEMENTS = "Valet iOS Test Host App/Valet iOS Test Host App.entitlements";
+				CODE_SIGN_IDENTITY = "Apple Development";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				CODE_SIGN_STYLE = Automatic;
 				DEBUG_INFORMATION_FORMAT = dwarf;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = 9XUJ7M53NG;
```
NickEntin, line 2527: Where did this development team ID come from?

Sources/Valet/SharedGroupIdentifier.swift
```diff
@@ -35,23 +35,47 @@ public struct SharedAccessGroupIdentifier: CustomStringConvertible {
...
 
-        self.appIDPrefix = appIDPrefix
+        self.prefix = appIDPrefix
+        self.groupIdentifier = groupIdentifier
+    }
+
+    /// A representation of a shared app group identifier.
+    /// - Parameters:
+    ///   - groupPrefix: On iOS, iPadOS, watchOS, and tvOS, this prefix must equal "group". On macOS, this prefix is the application's App ID prefix, which  can be found by inspecting the application's provisioning profile, or viewing the application's App ID Configuration on developer.apple.com. This string must not be empty.
+    ///   - groupIdentifier: An identifier that cooresponds to a value in com.apple.security.application-groups in the application's Entitlements file. This string must not be empty.
```
NickEntin:
```suggestion
    ///   - groupIdentifier: An identifier that corresponds to a value in com.apple.security.application-groups in the application's Entitlements file. This string must not be empty.
```

Sources/Valet/SharedGroupIdentifier.swift
```diff
@@ -35,23 +35,47 @@ public struct SharedAccessGroupIdentifier: CustomStringConvertible {
...
+    /// - SeeAlso: https://developer.apple.com/documentation/security/keychain_services/keychain_items/sharing_access_to_keychain_items_among_a_collection_of_apps
+    public init?(groupPrefix: String, nonEmptyGroup groupIdentifier: String?) {
```
NickEntin, line 47: Should we enforce the group prefix on non-macOS by having separate inits?

```swift
#if os(macOS)
public init?(groupPrefix: String, nonEmptyGroup groupIdentifier: String?) {
#else
public init?(nonEmptyGroup groupIdentifier: String?) {
    let groupPrefix = Self.appGroupPrefix
#endif
```

NickEntin, line 47: Does `groupIdentifier` need to be optional? Given the non-empty requirement, it feels like we should also make it non-optional.

Sources/Valet/SharedGroupIdentifier.swift
dfed, line 45:
```suggestion
    ///   - groupIdentifier: An identifier that corresponds to a value in com.apple.security.application-groups in the application's Entitlements file. This string must not be empty.
```

Valet.xcodeproj/project.pbxproj
dfed: This is the valet-open-source@squareup.com identifier. This likely shouldn't have been committed.

README.md
```diff
@@ -125,17 +125,29 @@ NSString *const myLuggageCombination = [myValet stringForKey:username error:nil]
...
+
+```swift
+let mySharedValet = Valet.sharedGroupValet(with: SharedGroupIdentifier(groupPrefix: "group", nonEmptyGroup: "Druidia")!, accessibility: .whenUnlocked)
+```
+
+```objc
+VALValet *const mySharedValet = [VALValet sharedGroupValetWithGroupPrefix:@"group" sharedGroupIdentifier:@"Druidia" accessibility:VALAccessibilityWhenUnlocked];
+```
+
+This instance can be used to store and retrieve data securely across any app written by the same developer that has `group.Druidia` set as a value for the `com.apple.security.application-groups` key in the app’s `Entitlements`. This Valet is accessible when the device is unlocked. Note that `myValet` and `mySharedValet` can not read or modify one another’s values because the two Valets were created with different initializers. All Valet types can share secrets across applications written by the same developer by using the `sharedGroupValet` initializer. Note that on macOS, the `groupPrefix` [must be the App ID prefix](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_application-groups#discussion).
```
NickEntin, line 150:
```suggestion
This instance can be used to store and retrieve data securely across any app written by the same developer that has `group.Druidia` set as a value for the `com.apple.security.application-groups` key in the app’s `Entitlements`. This Valet is accessible when the device is unlocked. Note that `myValet` and `mySharedValet` cannot read or modify one another’s values because the two Valets were created with different initializers. All Valet types can share secrets across applications written by the same developer by using the `sharedGroupValet` initializer. Note that on macOS, the `groupPrefix` [must be the App ID prefix](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_application-groups#discussion).
```

Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
```diff
@@ -159,7 +159,24 @@ class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
             XCTAssertTrue(permutation.canAccessKeychain())
         }
     }
-    
+
+    func test_canAccessKeychain_sharedAppGroup() {
+        guard #available(tvOS 11.0, *) else {
+            return
+        }
```
NickEntin, line 168: Instead of guarding, can we mark the test as `@available`?

```suggestion
    @available(tvOS 11.0, *)
    func test_canAccessKeychain_sharedAppGroup() {
```

Sources/Valet/SharedGroupIdentifier.swift
dfed: We'd need to put the entire initializer in the `#if` (`#if` in Swift is way less permissive than Objective-C), but that'd be doable. I can separate that out.

dfed: We made it optional in the other method above. I think it was a convenience, similar to how `Identifier` works. Open to changing this everywhere. LMK how you're thinking about this.

Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
dfed: unfortunately no. I learned the hard way in earlier Valet PRs that `@available` doesn't work as expected when applied to test methods – all methods are executed, regardless of their availability. I believe this is because XCTest uses the Objective-C runtime to find all selectors that begin with `test`.

Valet.xcodeproj/project.pbxproj
dfed: fa682b14a03884d6781930015d746332af65004d

Sources/Valet/SharedGroupIdentifier.swift
dfed: Actually, looking at this more: I think having the `groupPrefix` makes it clear what the `groupIdentifier` should be. If I make a single-parameter initializer for non-macOS platforms, then I think I'd need to accept `groupIdentifier`'s that have `group.` as a prefix, as well as ones that do not. I Kina like this API being explicitly not magical. Thoughts?

NickEntin: I think making the `groupIdentifier` only take the non-prefix string is fine as long as we're clear in the headerdoc what the parameter expects (i.e. "this is the identifier following the 'group.' in the entitlement").

I agree with your point that having the `groupPrefix` makes it more obvious, but I think that's outweighed by having a parameter that only accepts a single string. Don't feel strongly though.

NickEntin: That's fine, happy to match the existing API.

NickEntin: If we switch to separate initializers, we could also clean up the macOS case by removing the concept of a "group prefix" and specifying that it's the app id prefix.

```swift
init?(nonEmptyGroup groupIdentifier: String?)
init?(appIDPrefix: String, nonEmptyGroup groupIdentifier: String?)
```

dfed: I like that! But the tough thing is we already have this initializer for shared access groups:
```
init?(appIDPrefix: String, nonEmptyGroup groupIdentifier: String?)
```

There's a good bit of overlap between how these features work. I'm leaning hard towards keeping the APIs explicit for now. We can always add new convenience initializers down the line.

Issue #231, opened 2020-05-24
Can't find object after switching accessibility


macrossyun:
Recently we changed the accessibility from **.whenUnlockedThisDeviceOnly** to **.afterFirstUnlockThisDeviceOnly**, and it is causing the object we stored in the keychain can not be found after users update the app.

We stored the id token in the keychain, and it force the users to log out the app. However after they login back, the problem just doesn't happen anymore. 

The version of Valet we are using is 3.2.5

dfed:
Thanks for the report @macrossyun!

Can you help me understand the steps you took? It sounds like you changed the accessibility on your Valet, but did you migrate the values from the old Valet configuration to the new Valet configuration using the `migrateObjects(from:removeOnCompletion:)` method?

Our README states that:
> Valets created with a different class type, via a different initializer, or with a different accessibility attribute will not be able to read or modify values in `myValet`.

If you didn’t migrate after changing accessibility values, please let me know. I can re-task this issue to track making an explicit section in the README on “Changing your Accessibility value” to try to make this workflow more clear.

macrossyun:
Thanks for your reply @dfed. Yeah, I think that must be the reason, I didn't notice that line when I changed the accessibility. As I used to use other library before with similar setting, I thought it wouldn't be the problem when change the accessibility........

Sorry my bad

dfed:
Thanks for letting me know! I’ll update the README to make this more clear.

macrossyun:
Thanks @dfed !

PR #232, opened 2020-05-25
Create 'Changing an Accessibility Value After Persisting Data' section in README

dfed:
Hopefully this section will help our customers avoid the trap laid bare in #231. I think being more explicit about the Valet sandboxing behavior earlier in the README is a good thing as well. Prior to this PR, the sandboxing behavior was acknowledged but pretty buried.

cc @macrossyun please let me know if this is clear.

I also fixed a copy/paste issue in the README while I was here: we were using `mySecureEnclaveValet` in an example before we'd discussed what a Secure Enclave Valet was.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ VALValet *const myNewValet = [VALValet valetWithExplicitlySetIdentifier:@"Druidi
 [myNewValet migrateObjectsFrom:myOldValet removeOnCompletion:true error:nil];
 ```
 
-The Valet type, identifier, accessibility value, and initializer chosen to create a Valet are combined to create a sandbox within the keychain. This behavior ensures that different Valets can not read or write one another's key:value pairs. If you change a Valet's accessibility after persisting key:value pairs, you must migrate the key:values pairs from the Valet with the no-longer-desired accessibility to the Valet with the desired accessibility to avoid data loss.
+The Valet type, identifier, accessibility value, and initializer chosen to create a Valet are combined to create a sandbox within the keychain. This behavior ensures that different Valets can not read or write one another's key:value pairs. If you change a Valet's accessibility after persisting key:value pairs, you must migrate the key:value pairs from the Valet with the no-longer-desired accessibility to the Valet with the desired accessibility to avoid data loss.
 
 ### Reading and Writing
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,6 +109,22 @@ Mac apps signed with a developer ID may see their Valet’s identifier [shown to
...
+try? myNewValet.migrateObjects(from: myOldValet, removeOnCompletion: true)
+```
+
+```objc
+VALValet *const myOldValet = [VALValet valetWithExplicitlySetIdentifier:@"Druidia" accessibility:VALAccessibilityWhenUnlocked];
+VALValet *const myNewValet = [VALValet valetWithExplicitlySetIdentifier:@"Druidia" accessibility:VALAccessibilityAfterFirstUnlock];
+[myNewValet migrateObjectsFrom:myOldValet removeOnCompletion:true error:nil];
+```
+
+The Valet type, identifier, accessibility value, and initializer chosen to create a Valet are combined to create a sandbox within the keychain. This behavior ensures that different Valets can not read or write one another's key:value pairs. If you change a Valet's accessibility after persisting key:value pairs, you must migrate the key:values pairs from the Valet with the no-longer-desired accessibility to the Valet with the desired accessibility to avoid data loss.
```
NickEntin, line 126:
```suggestion
The Valet type, identifier, accessibility value, and initializer chosen to create a Valet are combined to create a sandbox within the keychain. This behavior ensures that different Valets can not read or write one another's key:value pairs. If you change a Valet's accessibility after persisting key:value pairs, you must migrate the key:value pairs from the Valet with the no-longer-desired accessibility to the Valet with the desired accessibility to avoid data loss.
```
`key:values` -> `key:value`

NickEntin: total nit: Should this paragraph come before the code samples, so we explain what's going on first?

dfed: > Should this paragraph come before the code samples, so we explain what's going on first?

Possibly. I was following the examples in the other sections where the code snippet comes first. I don't really want to re-think that system right now 😅, so I'll leave it as-is.

macrossyun:
> Hopefully this section will help our customers avoid the trap laid bare in #231. I think being more explicit about the Valet sandboxing behavior earlier in the README is a good thing as well. Prior to this PR, the sandboxing behavior was acknowledged but pretty buried.
> 
> cc @macrossyun please let me know if this is clear.
> 
> I also fixed a copy/paste issue in the README while I was here: we were using `mySecureEnclaveValet` in an example before we'd discussed what a Secure Enclave Valet was.

👍, thanks @dfed

PR #237, opened 2020-06-22
Use XCTSkipUnless

dfed:
@fdiaz made me aware of [Xcode's built-in ability to skip tests](https://developer.apple.com/documentation/xctest/skipping_tests). It's a more semantic way to skip a test when conditions for running the test aren't met. Previously we were using `guard`, but since there's a more semantic call we can use, let's use it!


dfed:
Shoot. Looks like the minimum Xcode target for this API is Xcode 11.4. Closing this PR. Might be time to start a Valet 5 wishlist 😅

fdiaz:
Yup! Looks like it https://developer.apple.com/documentation/xctest/xctskip

Thanks for trying though!

PR #247, opened 2020-09-14
Make it clear that Valet is not meant to store large data blobs

### LINKED ISSUE #246: Error: cannotAccessKeychain despite `canAccessKeychain()` returning true immediately before calling, caused by saving largeish data
I've ran into a bug where saving to the Keychain with Valet will fail with a "cannotAccessKeychain" error if you're saving larger amounts of data (say, several megabytes) despite everything else seeming proper and `canAccessKeychain()`  being called immediately before returning `true`.

I ran into this because I foolishly was testing encrypting more of the user's data (performance was fine on modern devices) than necessary, and for some users with a large amount of data combined with older iOS devices they were running into issues with data saving. Sure enough I was able to replicate this within a few tries with an iPhone 6 running iOS 12.4.3.

I was admittedly lulled to a false sense of security a little (totally my fault, should have done my due diligence) by this phrase in the README that maybe should be updated to reflect issues around larger data saving:

> Valet guarantees it will never fail to write to or read from the keychain unless canAccessKeychain() returns false. There are only a few cases that can lead to the keychain being inaccessible:

In this can `canAccessKeychain` indeed returns true, but the call immediately following it will fail. Here's an exaggerated test case showing it with a 10MB data set (excuse the forced unwrapping, just for illustrative purposes):

```swift
var testDict: [String: String] = [:]
        
for i in 0 ..< 100_000 {
    testDict["\(i)"] = "once upon a time there was a ghost named earl the ghost"
 }

let largeData = try! NSKeyedArchiver.archivedData(withRootObject: testDict, requiringSecureCoding: true)
print(largeData.count) // 10457973 bytes, or 10.4MB

let valet = Valet.sharedGroupValet(with: SharedGroupIdentifier(appIDPrefix: "---", nonEmptyGroup: "---")!, accessibility: .afterFirstUnlock)

// Just to show that we're in pretty much the optimal situation
if valet.canAccessKeychain() && UIApplication.shared.applicationState == .active && UIApplication.shared.isProtectedDataAvailable {
    let smallData = "foo".data(using: .utf8)!
    try! valet.setObject(smallData, forKey: "test123") // Success!
    try! valet.setObject(largeData, forKey: "test456") // Error: cannot access keychain
    print("Complete! But we often don't reach this point.")
}
```

(It may take a few tries to reproduce, but I see the same error that users are reporting. On main thread in this example if relevant, but occurs on background threads as well. Also my logs show the error occurring on the most up to date version of iOS 13 (13.7) as well as the most recent iOS 14 beta, so it doesn't seem to be limited to iOS 12. Also worth mentioning it happens in other libraries like [Lockbox](https://github.com/granoff/Lockbox) so it appears to be an internal Keychain issue, not specific to Valet, but still might be worth mentioning in that exceptions section.)

Again, 100% my dumb fault for saving things larger than I likely should be and not testing thoroughly enough on older devices, but thought it might be helpful to mention in case anyone else runs into it.

And thank you for the library!

dfed:
Resolves #246


README.md
```diff
@@ -222,11 +222,12 @@ If your macOS application supports macOS 10.14 or prior, you must run `myValet.m
 Valet guarantees it will never fail to write to or read from the keychain unless `canAccessKeychain()` returns `false`. There are only a few cases that can lead to the keychain being inaccessible:
 
 1. Using the wrong `Accessibility` for your use case. Examples of improper use include using `.whenPasscodeSetThisDeviceOnly` when there is no passcode set on the device, or using `.whenUnlocked` when running in the background.
-2. Initializing a Valet with shared access group Valet when the shared access group identifier is not in your entitlements file.
```
dfed, line 225: I changed all the numbers here to `1.` to make this list easier to maintain.

README.md
```diff
@@ -222,11 +222,12 @@ If your macOS application supports macOS 10.14 or prior, you must run `myValet.m
...
+1. Running your app on device or in the simulator with a debugger attached may also [cause an entitlements error](https://forums.developer.apple.com/thread/4743) to be returned when reading from or writing to the keychain. To work around this issue on device, run the app without the debugger attached. After running once without the debugger attached the keychain will usually behave properly for a few runs with the debugger attached before the process needs to be repeated.
+1. Running your app or unit tests without the application-identifier entitlement. Xcode 8 introduced a requirement that all schemes must be signed with the application-identifier entitlement to access the keychain. To satisfy this requirement when running unit tests, your unit tests must be run inside of a host application.
+1. Attempting to write data more than 1mb in size. The Keychain is built to securely store small secrets – writing large blobs can crash Apple's Security daemon.
```
dfed, line 230: This bullet is the only new item in this list.

README.md
```diff
@@ -222,11 +222,12 @@ If your macOS application supports macOS 10.14 or prior, you must run `myValet.m
 Valet guarantees it will never fail to write to or read from the keychain unless `canAccessKeychain()` returns `false`. There are only a few cases that can lead to the keychain being inaccessible:
```
NickEntin, line 222: nit: Maybe update this to "... will never fail to write _valid data_ to or read from..." to indicate that there are certain data inputs where it still might fail?

NickEntin (approved):

README.md
dfed: Solid call. This also covers the case where someone tries to write an empty string to the keychain.
```suggestion
Valet guarantees it will never fail to read from or write valid data to the keychain unless `canAccessKeychain()` returns `false`. There are only a few cases that can lead to the keychain being inaccessible:
```

PR #248, opened 2020-09-14
Recommend that data should be no longer than REASONABLE_SECRET_DATA_SIZE

dfed:
See [this file](https://opensource.apple.com/source/Security/Security-59306.120.7/keychain/securityd/SecDbKeychainItemV7.m) for the constant. Follow up to #246 and #247.

Given that we have the known upper desired limit of keychain secret size (thank you @KhaosT!), I think it's reasonable for us to create a new `KeychainError` case `valueTooLarge`, and check that the entered data isn't above that limit before interacting with the `securityd`. If and when we do that, we'd want to change this `/// - Important` to a `/// - Precondition`.

Open to someone putting out a PR that does that (or I can get to it at a future date). We'd want to do this check [within `Keychain`'s `setObject(_:forKey:options:)`](https://github.com/square/Valet/blob/ec9c2a3999d2c3e2626c134b0155395ce1c58ce6/Sources/Valet/Internal/Keychain.swift#L91).

But for now, let's make sure our documentation is telling our customers the right thing per https://github.com/square/Valet/issues/246#issuecomment-691794497.

### DIFF
--- a/Sources/Valet/SecureEnclaveValet.swift
+++ b/Sources/Valet/SecureEnclaveValet.swift
@@ -119,7 +119,7 @@ public final class SecureEnclaveValet: NSObject {
     ///   - object: A Data value to be inserted into the keychain.
     ///   - key: A key that can be used to retrieve the `object` from the keychain.
     /// - Throws: An error of type `KeychainError`.
-    /// - Important: Inserting data larger than 1mb in size may sporadically fail.
+    /// - Important: Inserted data should be smaller than 4kb.
     @objc
     public func setObject(_ object: Data, forKey key: String) throws {
         try execute(in: lock) {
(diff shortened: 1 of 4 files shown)


Sources/Valet/SecureEnclaveValet.swift
```diff
@@ -119,7 +119,7 @@ public final class SecureEnclaveValet: NSObject {
     ///   - object: A Data value to be inserted into the keychain.
     ///   - key: A key that can be used to retrieve the `object` from the keychain.
     /// - Throws: An error of type `KeychainError`.
-    /// - Important: Inserting data larger than 1mb in size may sporadically fail.
+    /// - Important: Inserted data should be smaller than 4kb.
```
NickEntin, line 122: supernit: It looks like it supports _up to_ 4kb, rather than requiring it be smaller than that.
```suggestion
    /// - Important: Inserted data should be no larger than 4kb.
```

NickEntin (approved):

NickEntin:
> create a new `KeychainError` case `keyTooLarge`

I think the `REASONABLE_SECRET_DATA_SIZE` is referring to the value, not the key. Or maybe both? Might be worth adding both `keyTooLarge` and `valueTooLarge` errors?

dfed:
@NickEntin 🤦‍♂️ I def meant `valueTooLarge` not `keyTooLarge`. Good catch. I've updated the description.

Regarding key size: I haven't found any explicit key-size comparisons in the linked code. We can cross that bridge if we find it.

PR #255, opened 2021-02-11
Add convenience methods for `Codable` types

### LINKED ISSUE #254: Add convenience access methods for `Codable` types
As of now, we have basically two methods for accessing data, namely `string` (`string(forKey:)`, `setString(:forKey:)`) and `object` (`object(forKey:)`, `setObject(:forKey:)`). While this covers all use cases as users can convert any type to `Data` and store it via `object` to the keychain, it's quite inconvenient to write the conversion code each time. Especially now that Swift has an official way of specifying that types can be made `Codable` to communicate to others (incl. this library) that there's a uniform way of archiving and unarchiving the type.

So what I suggest is that we add convenience methods to this great library which automate the conversion to and from `Data` for all `Codable` types. To simplify the code here in the library, we probably shouldn't take any responsibility of reacting to conversion errors, instead we should just rethrow them and let the users handle any issues with encoding/decoding. For that I think it's important that we also document as part of the new APIs which encoder/decoder we're using so that users of this library can transparently handle the possible errors.

Jeehut:
This fixes #254.

I also added some basic tests, but not as detailed because these new methods just call other already well tested methods. I didn't see the need to re-test their logic here as that would only create redundancy. But I added tests for the rethrowing behavior of decoding errors so the library doesn't take any responsibility for error handling there.

Possible changelog entry (couldn't find a `CHANGELOG.md` file):
- Adds new `codable(forKey:jsonDecoder:)`, `setCodable(:forKey:jsonEncoder:)` and `containsCodable(forKey:)` convenience methods for storing `Codable` types. Defaults to `JSONEncoder()`/`JSONDecoder()` for convenience.

Please note that I have also made **2 minor adjustments** to improve the code elsewhere (which I can remove if requested):
1. I renamed the comment in the tests from `// MARK: set(object:forKey:)` to `// MARK: setObject(:forKey:)` to use the correct method name.
2. Added explicit project preference settings for `tabWidth` and `indentWidth` so contributors using a different global setting don't mess with indentation accidentally.

### DIFF
--- a/Sources/Valet/Valet.swift
+++ b/Sources/Valet/Valet.swift
@@ -285,29 +285,29 @@ public final class Valet: NSObject {
         }
     }
 
-    /// Convenience setter for `Codable` objects. Stores them using `JSONEncoder` and the `setObject(:forKey:)` method.
+    /// Convenience setter for `Codable` objects. Utilizes the passed-in encoder to encode before storing the value in the Keychain with the `setObject(:forKey:)` method.
     ///
     /// - Parameters:
     ///   - object: A `Codable` object to be inserted into the keychain.
     ///   - key: A key that can be used to retrieve the `object` from the keychain.
-    ///   - jsonEncoder: The `JSONEncoder` to use for encoding the type to store. Defaults to the default initializer.
+    ///   - encoder: The `JSONEncoder` to use for encoding the type to store. Defaults to the default initializer.
     /// - Throws: An error of type `KeychainError` or an error produced while trying to encode via `JSONEncoder`.
     /// - Important: Inserted objects JSON data representation should be no larger than 4kb.
-    public func setCodable<T: Codable>(_ object: T, forKey key: String, jsonEncoder: JSONEncoder = .init()) throws {
-        let jsonData = try jsonEncoder.encode(object)
-        try setObject(jsonData, forKey: key)
+    public func setCodable<T: Codable>(_ object: T, forKey key: String, encoder: JSONEncoder = .init()) throws {
+        let encodedData = try encoder.encode(object)
+        try setObject(encodedData, forKey: key)
     }
 
-    /// Convenience getter for `Codable` objects. Reads them using `JSONDecoder` and the `object(forKey:)` method.
+    /// Convenience getter for `Codable` objects. Utilizes the passed-in decoder to decode after reading the value from the Keychain with the `object(forKey:)` method.
     ///
     /// - Parameters:
     ///   - key: A key used to retrieve the desired object from the keychain.
-    ///   - jsonDecoder: The `JSONDecoder` to use for decoding the stored data. Defaults to the default initializer.
+    ///   - decoder: The `JSONDecoder` to use for decoding the stored data. Defaults to the default initializer.
     /// - Returns: The decoded object currently stored in the keychain for the provided key.
     /// - Throws: An error of type `KeychainError` or an error produced while trying to decode via `JSONDecoder`.
-    public func codable<T: Codable>(forKey key: String, jsonDecoder: JSONDecoder = .init()) throws -> T {
-        let jsonData = try object(forKey: key)
-        return try jsonDecoder.decode(T.self, from: jsonData)
+    public func codable<T: Codable>(forKey key: String, decoder: JSONDecoder = .init()) throws -> T {
+        let encodedData = try object(forKey: key)
+        return try decoder.decode(T.self, from: encodedData)
     }
 
     /// Convenience method name for `containsObject(forKey:)` for API completeness.
--- a/Sources/Valet/Valet.swift
+++ b/Sources/Valet/Valet.swift
@@ -285,6 +285,37 @@ public final class Valet: NSObject {
         }
     }
 
+    /// Convenience setter for `Codable` objects. Stores them using `JSONEncoder` and the `setObject(:forKey:)` method.
+    ///
+    /// - Parameters:
+    ///   - object: A `Codable` object to be inserted into the keychain.
+    ///   - key: A key that can be used to retrieve the `object` from the keychain.
+    /// - Throws: An error of type `KeychainError` or an error produced while trying to encode via `JSONEncoder`.
+    /// - Important: Inserted objects JSON data representation should be no larger than 4kb.
+    public func setCodable<T: Codable>(_ object: T, forKey key: String) throws {
+        let jsonData = try JSONEncoder().encode(object)
+        try setObject(jsonData, forKey: key)
+    }
+
+    /// Convenience getter for `Codable` objects. Reads them using `JSONDecoder` and the `object(forKey:)` method.
+    ///
+    /// - Parameter key: A key used to retrieve the desired object from the keychain.
+    /// - Returns: The decoded object currently stored in the keychain for the provided key.
+    /// - Throws: An error of type `KeychainError` or an error produced while trying to decode via `JSONDecoder`.
+    public func codable<T: Codable>(forKey key: String) throws -> T {
+        let jsonData = try object(forKey: key)
+        return try JSONDecoder().decode(T.self, from: jsonData)
+    }
+
+    /// Convenience method name for `containsObject(forKey:)` for API completeness.
+    ///
+    /// - Parameter key: The key to look up in the keychain.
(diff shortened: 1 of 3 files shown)


dfed:
Thanks for being so proactive! Let's align on some details on the linked issue before I review this PR.

Regarding the lack of a checked-in changelog: this repo stores release information in GitHub's releases UI. I appreciate your writing up a draft release note! That'll save me some time.

Also: TIL about project settings for tab width. I'll need to take that learning to my other projects!

Jeehut:
@dfed I just adjusted the PR with a new `jsonEncoder`/`jsonDecoder` parameter, the rationale is explained in #254.

Yeah, took me a while as well to learn that we can explicitly specify the indentation preferences per project although it seems to be a very old Xcode feature. Reason I changed it by the way was very selfish: Since I switched to Apples [Swift-Format](https://github.com/apple/swift-format) (from SwiftLint) which prefers 2 whitespaces by default I have my global settings set to 2 spaces, thus it helped me while developing on this PR. :)

Jeehut:
Hmm, tests with the new parameter seem to be failing, at least in Xcode 13.3. I'm going to download 11.3 to check what the issue is, for me all tests pass locally. But Xcode 13.3 was released on 10 Dec 2019 which is over 2 years old, so maybe it's time for an update anyway?

dfed:
> Hmm, tests with the new parameter seem to be failing, at least in Xcode 11.3. I'm going to download 11.3 to check what the issue is, the CI errors don't really help and for me all tests pass locally.

Shoot. I'd forgotten we were supporting Xcode 11 still. Dropping an Xcode version is a major version bump, so we won't be able to drop support for Xcode 11. That may force our hand in terms of how we solve the generic encoder/decoder problem. Our CI is intended to be set to our lowest-supported Xcode version specifically to help us catch issues like this.

Sources/Valet/Valet.swift
```diff
@@ -285,6 +285,40 @@ public final class Valet: NSObject {
         }
     }
 
+    /// Convenience setter for `Codable` objects. Stores them using `JSONEncoder` and the `setObject(:forKey:)` method.
```
dfed, line 288:
```suggestion
    /// Convenience setter for `Codable` objects. Utilizes the passed-in encoder to encode before storing the value in the Keychain with the `setObject(:forKey:)` method.
```

Sources/Valet/Valet.swift
```diff
@@ -285,6 +285,40 @@ public final class Valet: NSObject {
...
+    ///
+    /// - Parameters:
+    ///   - object: A `Codable` object to be inserted into the keychain.
+    ///   - key: A key that can be used to retrieve the `object` from the keychain.
+    ///   - jsonEncoder: The `JSONEncoder` to use for encoding the type to store. Defaults to the default initializer.
```
dfed, line 293: Once we allow a generic encoder to be passed in, we'll need to change this comment and parameter name.

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -451,7 +452,7 @@ class ValetIntegrationTests: XCTestCase
         }
     }
     
-    // MARK: set(object:forKey:)
+    // MARK: setObject(:forKey:)
```
dfed, line 455: thank you for fixing this!

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -481,6 +482,57 @@ class ValetIntegrationTests: XCTestCase
...
+        try allPermutations.forEach { valet in
+            try valet.setCodable(stringArray, forKey: key)
+            XCTAssertThrowsError(try valet.codable(forKey: key) as [Int]) { error in
+                switch error {
+                case DecodingError.typeMismatch:
+                    break // expected
+
+                default:
+                    XCTFail("Unexpected error: \(error)")
+                }
```
dfed, line 505: I think the following will compile and will be a little more concise without losing readability:
```suggestion
                XCTAssertEquals(error as? DecodingError, .typeMismatch)
```

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -481,6 +482,57 @@ class ValetIntegrationTests: XCTestCase
...
+
+    func test_setCodableForKey_throwsEmptyKeyOnInvalidKey() throws {
+        try allPermutations.forEach { valet in
+            XCTAssertThrowsError(try valet.setCodable(stringArray, forKey: "")) { error in
+                XCTAssertEqual(error as? KeychainError, .emptyKey)
+            }
+        }
+    }
+
+    func test_setCodableForKey_customCoders() throws {
```
dfed, line 520: Let's elaborate about what we're testing with `customCoders`. Maybe (and feel free to correct if this doesn't adequately capture your intent):
```suggestion
    func test_setCodableForKey_utilizesPassedInCoder() throws {
```

Valet.xcodeproj/project.pbxproj
```diff
@@ -836,7 +836,9 @@
 				EA1E1F841A8C46080067C991 /* Products */,
 				168909331F7199B50057F636 /* Frameworks */,
 			);
+			indentWidth = 4;
 			sourceTree = "<group>";
+			tabWidth = 4;
```
dfed, line 841: Absolutely love this. Thanks for sharing 🙂

dfed (commented):
This is a great start! I'd love to:
* Enable using coders other than JSON
* Add these convenience methods to `SecureEnclaveValet` and `SinglePromptSecureEnclaveValet` as well

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
Jeehut: This suggestion did not work, I'm getting 2 errors:
1. `Cannot convert value of type '(Any.Type, DecodingError.Context) -> DecodingError' to expected argument type 'DecodingError?'`
2. `Global function 'XCTAssertEqual(_:_:_:file:line:)' requires that 'DecodingError' conform to 'Equatable'`

Jeehut:
@dfed Thank you for the code review & suggestions. I tried to fix them all except for making this work with non-JSON coders & adding same API to the other valet types. I'm personally not using those APIs and have limited knowledge about them, thus I'm probably not the right person to write the code there, also I don't have the time.

I tried making the functions generic using this approach:

```swift
#if canImport(Combine)
import Combine
#endif

[...]

    #if canImport(Combine)
    @available(iOS 13.0, macOS 10.15, watchOS 6.0, *)
    public func setCodable<T: Codable, E: TopLevelEncoder>(_ object: T, forKey key: String, encoder: E) throws where E.Output == Data {
        let encodedData = try encoder.encode(object)
        try setObject(encodedData, forKey: key)
    }

    @available(iOS 13.0, macOS 10.15, watchOS 6.0, *)
    public func codable<T: Codable, D: TopLevelDecoder>(forKey key: String, decoder: D) throws -> T where D.Input == Data {
        let encodedData = try object(forKey: key)
        return try decoder.decode(T.self, from: encodedData)
    }
    #endif
```

But then I couldn't specify a default encoder & decoder, so I needed to add my methods from the PR as it is right now anyways and I didn't want to add too much on the API. I personally think that most users of Valet will not care much about how the data is stored internally as long as Valet can decode it again, so restricting to `JSONEncoder`/`JSONDecoder` for now but keeping the code extensible (with above methods) should work.

But please note that the CI still seems to be failing, even with your `CodableArray` struct suggestion. Or did I do it the wrong way? I couldn't test on Xcode 11.3 yet as I didn't have the time to download & install it yet.

thedavidharris:
> Edit: I just checked what the CI error was. If I recall Xcode 11 had trouble encoding top-level objects. Instead of encoding an `Array<String>` directly, instead make a type that can be encoded that has an `Array<String>` like follows:

https://bugs.swift.org/browse/SR-6163 was the bug, so would have to work around it as long as iOS 12 is supported iirc

dfed:
Welp – this fell off my to-do list. Apologies for the severe delay here.

@Jeehut I appreciate the effort you've put into this, but if I'm reading the situation right I don't think either of us have the time to get this PR over the line. If I'm right about that, let's close this one out. Hopefully this discussion and sample PR will help the next person who tries to take this up.

PR #259, opened 2021-02-15
Document our ability to unlock with watch

### LINKED ISSUE #253: Feature Request, Unlock with Apple Watch
```swift
@available(macCatalyst 13.0, *)
@available(iOS, unavailable)
public static var watch: SecAccessControlCreateFlags { get }
```

more info [here](https://developer.apple.com/documentation/security/secaccesscontrolcreateflags/3042482-watch)

dfed:
Resolves #253

@plananderson

### DIFF
--- a/Sources/Valet/SecureEnclaveAccessControl.swift
+++ b/Sources/Valet/SecureEnclaveAccessControl.swift
@@ -19,7 +19,7 @@ import Foundation
 
 @objc(VALSecureEnclaveAccessControl)
 public enum SecureEnclaveAccessControl: Int, CustomStringConvertible, Equatable {
-    /// Access to keychain elements requires user presence verification via Touch ID, Face ID, or device Passcode. On macOS, this element may also be accessed via a prompt on a paired watch. Keychain elements are still accessible by Touch ID even if fingers are added or removed. Touch ID does not have to be available or enrolled.
+    /// Access to keychain elements requires user presence verification via Touch ID, Face ID, or device Passcode. On macOS 10.15 and later, this element may also be accessed via a prompt on a paired watch. Keychain elements are still accessible by Touch ID even if fingers are added or removed. Touch ID does not have to be available or enrolled.
     case userPresence = 1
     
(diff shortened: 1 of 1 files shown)


Sources/Valet/SecureEnclaveAccessControl.swift
```diff
@@ -19,10 +19,10 @@ import Foundation
...
+    /// Access to keychain elements requires user presence verification via Touch ID, Face ID, or device Passcode. On macOS, this element may also be accessed via a prompt on a paired watch. Keychain elements are still accessible by Touch ID even if fingers are added or removed. Touch ID does not have to be available or enrolled.
     case userPresence = 1
     
-    /// Access to keychain elements requires user presence verification via Face ID, or any finger enrolled in Touch ID. Keychain elements remain accessible via Face ID or Touch ID  after faces or fingers are added or removed. Face ID must be enabled with at least one face enrolled, or Touch ID must be available and at least one finger must be enrolled.
+    /// Access to keychain elements requires user presence verification via Face ID, or any finger enrolled in Touch ID. Keychain elements remain accessible via Face ID or Touch ID after faces or fingers are added or removed. Face ID must be enabled with at least one face enrolled, or Touch ID must be available and at least one finger must be enrolled.
```
dfed, line 25: Noticed the extra whitespace while I was here.

Sources/Valet/SecureEnclaveAccessControl.swift
```diff
@@ -97,18 +97,16 @@ public enum SecureEnclaveAccessControl: Int, CustomStringConvertible, Equatable
     }
 
     internal static func allValues() -> [SecureEnclaveAccessControl] {
```
dfed, line 99: There are no changes to the output of this method. However when investigating adding more cases I found that the previous setup of this method made it hard to add new cases with different `#available` configurations, so I restructured this method to be a bit easier to maintain over time.

sethfri (approved):
LGTM!

PR #262, opened 2021-03-30
Package.swift: adds .dynamic library

AndrewSB:


dfed:
@AndrewSB thank you for the contribution! I haven't shipped an app with SPM yet, so I'm a bit new to how SPM works with dynamic libraries. Ideally, it'd be possible for you to state at integration time in your app that you want a dynamic version of the Valet package we're providing. Is that only possible if we create a `.dynamic` library target?

From a little googling, it seems [possible to create your own package that vends a dynamic library that depends on our library](https://forums.swift.org/t/how-to-link-a-swift-package-as-dynamic/32062), though that's clearly not ideal.

I don't love the approach provided in this PR since it seems like we're papering over an issue with SPM + Xcode. But... if it is indeed the only way forward...

AndrewSB:
> state at integration time in your app that you want a dynamic version of the Valet package we're providing. Is that only possible if we create a .dynamic library target?

yeah, this is how it has been so far... not ideal.

but since making this, I've learned that Xcode 12.5b3 makes libraries dynamic if required automatically, if the `type` isn't specified.  lets leave this PR open for a couple days while i finish testing, and i'll respond here if i think this change is necessary

dfed:
Exciting that Xcode might fix this for us! Keep us posted 🙂

dfed:
Closing this out – if this PR is still relevant feel free to re-open and let us know the current status here.

AndrewSB:
it isn't! new Xcode totally handles this fine

PR #266, opened 2021-05-14
Log a clear message when keychain locks up due to the debugger being attached

efirestone:
This appears to be a bug in the Keychain, but hopefully this log message helps save some people time in the future.


efirestone:
Closing as this was caused by a different issue (my fault) and it's not clear that the original bug in the linked StackOverflow still exists after iOS 9.3.

PR #268, opened 2021-06-07
Fix failing tests

allisonmoyer:


Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
```diff
@@ -75,7 +75,7 @@ class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
         }
         
         try valet().setString(passcode, forKey: key)
-        let equivalentValet = SecureEnclaveValet.valet(with: valet().identifier, accessControl: .devicePasscode)
+        let equivalentValet = SinglePromptSecureEnclaveValet.valet(with: valet().identifier, accessControl: .devicePasscode)
```
dfed, line 78: This makes sense. Though maybe let's rename `equivalentValet` to `similarValet` since they aren't actually equivalent?

dfed (approved):
LGTM, but how was this test failing before? Def seems like a copy/paste issue but since we were testing that these were different I'd expect the tests to have been succeeding.

allisonmoyer:
> LGTM, but how was this test failing before? Def seems like a copy/paste issue but since we were testing that these were different I'd expect the tests to have been succeeding.

I'm not sure. There's more that are failing locally for me as well (like `test_secureEnclaveValetsWithDifferingAccessControl_canNotAccessSameData` and all of the `ValetBackwardsCompatibilityIntegrationTests` on macOS), so I was trying to trigger the test suite to see if my local failures would reflect on CI, but looks like they don't

dfed:
On target are the tests failing? If it's macOS there's likely a code-signing issue. Check out [these instructions](https://github.com/square/Valet/blob/master/Contributing.md#testing-changes-on-macos).

NickEntin (approved):

PR #269, opened 2021-06-08
Enable customization of biometrics prompt with application-specific fallback

### LINKED ISSUE #265: Enable customization of biometrics prompt with application-specific fallback
In order have more control when retrieving something secured by biometrics in the secure enclave, an option is to first create an `LAContext` and call `evaluatePolicy` with it, and then, on success, pass that authenticated context when looking up the item. However, this is not currently possible with the public interface to `SecureEnclaveValet`.

See more context [here](http://michael-brown.net/2018/touch-id-and-face-id-on-ios/) under "Restricting keychain item access to Touch ID and Face ID"

Proposition to fix is to add new public methods to `SecureEnclaveValet` to allow a context to be provided:
```
@objc
public func string(forKey key: String, withPrompt userPrompt: String, context: LAContext) throws -> String {
    try execute(in: lock) {
        keychainQuery[kSecUseAuthenticationContext as String] = context
        return try SecureEnclave.string(forKey: key, withPrompt: userPrompt, options: keychainQuery)
    }
}
```

and
```
@objc
public func object(forKey key: String, withPrompt userPrompt: String, context: LAContext) throws -> Data {
    try execute(in: lock) {
        keychainQuery[kSecUseAuthenticationContext as String] = context
        return try SecureEnclave.string(forKey: key, withPrompt: userPrompt, options: keychainQuery)
    }
}
```

allisonmoyer:
Resolves #265

### DIFF
--- a/Sources/Valet/SecureEnclaveValet.swift
+++ b/Sources/Valet/SecureEnclaveValet.swift
@@ -183,6 +183,7 @@ public final class SecureEnclaveValet: NSObject {
     ///                     If the user taps this button, an `SecureEnclaveError.userFallback` will be thrown.
     /// - Returns: The string currently stored in the keychain for the provided key.
     /// - Throws: An error of type `KeychainError` or `SecureEnclaveError`.
+    @objc
     public func string(
         forKey key: String,
         withPrompt userPrompt: String,
@@ -293,12 +294,9 @@ public final class SecureEnclaveValet: NSObject {
                         result = .failure(error)
                     }
 
-                } else if let error = error {
-                    result = .failure(SecureEnclaveError(error: error))
+                } else if let error = error as? LAError {
+                    result = .failure(SecureEnclaveError(error.code))
 
-                } else {
-                    // Unexpected to get here
-                    result = .failure(SecureEnclaveError.internalError)
                 }
 
                 semaphore.signal()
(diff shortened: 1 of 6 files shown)


Sources/Valet/SecureEnclaveValet.swift
```diff
@@ -226,13 +256,92 @@ public final class SecureEnclaveValet: NSObject {
...
+        withFallbackTitle fallbackTitle: String
+    ) throws -> String {
+        // Use a semaphore to block the thread and perform the evaluation synchronously.
+        let semaphore = DispatchSemaphore(value: 0)
+        var result: Result<String, Error>?
+        let policy = try accessControl.getPolicy()
+
+        let authenticationContext = authenticationContextProvider()
+        authenticationContext.localizedFallbackTitle = fallbackTitle
+        authenticationContext.evaluatePolicy(
```
allisonmoyer, line 294: I've spent some time over the past few days trying to use `evaluateAccessControl` here instead of `evaluatePolicy` so we don't have to try to map from our access control to `LAPolicy`, but `evaluateAccessControl` doesn't use the `localizedFallbackTitle` 😞

dfed: Welp. Good investigation at least!

PR #271, opened 2021-06-14
Enable customization of biometrics prompt with application-specific fallback

### LINKED ISSUE #265: Enable customization of biometrics prompt with application-specific fallback

allisonmoyer:
Resolves #265
Re-opens #267

### DIFF
--- a/Sources/Valet/SecureEnclaveError.swift
+++ b/Sources/Valet/SecureEnclaveError.swift
@@ -14,7 +14,7 @@
 //  limitations under the License.
 //
 
-import LocalAuthentication
+import Foundation
 
 @objc(VALSecureEnclaveError)
 public enum SecureEnclaveError: Int, CaseIterable, CustomStringConvertible, Error, Equatable {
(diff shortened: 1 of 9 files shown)


Sources/Valet/SecureEnclaveError.swift
```diff
@@ -0,0 +1,44 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+@objc(VALSecureEnclaveError)
+public enum SecureEnclaveError: Int, CaseIterable, CustomStringConvertible, Error, Equatable {
+
+    /// Access to the secure enclave was not attempted because the user opted to the fallback option (i.e. 'Enter Password').
```
NickEntin, line 22:
```suggestion
    /// Access to the secure enclave was not attempted because the user opted to use the fallback option (i.e. 'Enter Password').
```

NickEntin (approved):

allisonmoyer:
Thanks for all of the reviews @dfed @NickEntin. I'm now running into some availability issues for the `LAError.Code` cases. Have been looking at different ways to fix, but likely will just need to only make this function available for certain versions that support all of the code cases. Should have an update early next week.

Issue #270, opened 2021-06-11
Unit tests fail on tvOS 14


NickEntin:
Running on tvOS 14.5, we're getting a `-26276` response to `SecItemCopyMatching(_:_:)` in many of the `SinglePromptSecureEnclaveIntegrationTests` tests.

This doesn't match any of the `OSStatus` codes. The closest is `errSecDecode`, so maybe it's related to that?

```
errSecDecode                             = -26275,    /* Unable to decode the provided data. */
```

dfed:
Looking at other OSS Keychain code, it looks like `-26276` is an `internal` error, which isn't terribly helpful.

When I ran these tests locally, I saw the following error text in the console:
```
Valet tvOS Test Host App[70063:28557685] [Client,LAContext] LAContext[70063:0] failed to initialize: Error Domain=NSCocoaErrorDomain Code=4099 "The connection to service on pid 0 named com.apple.CoreAuthentication.daemon was invalidated." UserInfo={NSDebugDescription=The connection to service on pid 0 named com.apple.CoreAuthentication.daemon was invalidated.}
Valet tvOS Test Host App[70063:28557685] [Client,LAContext] LAContext[70063:0] failed to initialize: Error Domain=NSCocoaErrorDomain Code=4099 "The connection to service on pid 0 named com.apple.CoreAuthentication.daemon.libxpc was invalidated." UserInfo={NSDebugDescription=The connection to service on pid 0 named com.apple.CoreAuthentication.daemon.libxpc was invalidated.}
```

It seems like CoreAuthentication isn't working on tvOS 14, or at least it isn't working properly in the tvOS 14 simulators. It seems that prior to #256 we weren't testing tvOS 14 in CI, so it's quite likely this never worked.

I'm curious if physical tvOS devices work with `LAContext`s or `SinglePromptSecureEnclaveValet`s. I don't have a way to test this, however. Ideas?

dfed:
[...]

PR #284, opened 2022-08-12
Remove SinglePromptSecureEnclaveValet from tvOS targets

diogot:
LAContext is not available in tvOS, but for some reason it was compiling in older Xcode versions.

### DIFF
--- a/Sources/Valet/SinglePromptSecureEnclaveValet.swift
+++ b/Sources/Valet/SinglePromptSecureEnclaveValet.swift
@@ -14,6 +14,7 @@
 //  limitations under the License.
 //
 
+#if !os(tvOS)
 #if canImport(LocalAuthentication)
 
 import LocalAuthentication
@@ -21,7 +22,6 @@ import Foundation
 
 
 /// Reads and writes keychain elements that are stored on the Secure Enclave using Accessibility attribute `.whenPasscodeSetThisDeviceOnly`. The first access of these keychain elements will require the user to confirm their presence via Touch ID, Face ID, or passcode entry. If no passcode is set on the device, accessing the keychain via a `SinglePromptSecureEnclaveValet` will fail. Data is removed from the Secure Enclave when the user removes a passcode from the device.
-@available(tvOS 11.0, *)
 @objc(VALSinglePromptSecureEnclaveValet)
 public final class SinglePromptSecureEnclaveValet: NSObject {
     
@@ -277,7 +277,6 @@ public final class SinglePromptSecureEnclaveValet: NSObject {
 // MARK: - Objective-C Compatibility
 
 
-@available(tvOS 11.0, *)
 extension SinglePromptSecureEnclaveValet {
     
     // MARK: Public Class Methods
@@ -337,4 +336,5 @@ extension SinglePromptSecureEnclaveValet {
 
 }
 
+#endif
 #endif
--- a/Sources/Valet/SinglePromptSecureEnclaveValet.swift
+++ b/Sources/Valet/SinglePromptSecureEnclaveValet.swift
@@ -14,8 +14,8 @@
 //  limitations under the License.
 //
 
-#if !os(tvOS)
-#if canImport(LocalAuthentication)
+// Xcode 13 and prior incorrectly say that LocalAuthentication is available on tvOS, so we have to check both as long as Xcode 13 and prior are supported.
+#if !os(tvOS) && canImport(LocalAuthentication)
 
 import LocalAuthentication
 import Foundation
@@ -336,5 +336,4 @@ extension SinglePromptSecureEnclaveValet {
 
 }
 
-#endif
 #endif
--- a/Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
+++ b/Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
@@ -18,8 +18,8 @@ import Foundation
 @testable import Valet
 import XCTest
 
-#if !os(tvOS)
-#if canImport(LocalAuthentication)
+// Xcode 13 and prior incorrectly say that LocalAuthentication is available on tvOS, so we have to check both as long as Xcode 13 and prior are supported.
+#if !os(tvOS) && canImport(LocalAuthentication)
 
 class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
 {
@@ -235,5 +235,4 @@ class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
     }
 }
 
-#endif
 #endif
--- a/Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
+++ b/Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
@@ -18,13 +18,13 @@ import Foundation
 @testable import Valet
 import XCTest
 
+#if !os(tvOS)
 #if canImport(LocalAuthentication)
 
 class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
 {
     static let identifier = Identifier(nonEmpty: "valet_testing")!
 
-    @available(tvOS 11.0, *)
     func valet() -> SinglePromptSecureEnclaveValet {
         .valet(with: SinglePromptSecureEnclaveTests.identifier, accessControl: .userPresence)
     }
@@ -35,9 +35,6 @@ class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
     {
         super.setUp()
 
-        guard #available(tvOS 11.0, *) else {
-            return
-        }
         guard testEnvironmentIsSigned() else {
             return
         }
@@ -52,9 +49,6 @@ class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
         
     func test_SinglePromptSecureEnclaveValetsWithEqualConfiguration_canAccessSameData() throws
     {
-        guard #available(tvOS 11.0, *) else {
-            return
-        }
         guard testEnvironmentIsSigned() && testEnvironmentSupportsWhenPasscodeSet() else {
             return
         }
@@ -67,9 +61,6 @@ class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
     
     func test_SinglePromptSecureEnclaveValetsWithDifferingAccessControl_canNotAccessSameData() throws
     {
-        guard #available(tvOS 11.0, *) else {
-            return
-        }
         guard testEnvironmentIsSigned() && testEnvironmentSupportsWhenPasscodeSet() else {
             return
         }
@@ -87,9 +78,6 @@ class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
     
     func test_allKeys() throws
     {
-        guard #available(tvOS 11.0, *) else {
-            return
-        }
         guard testEnvironmentIsSigned() && testEnvironmentSupportsWhenPasscodeSet() else {
             return
         }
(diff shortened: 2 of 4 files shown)


NickEntin (approved):

Sources/Valet/SinglePromptSecureEnclaveValet.swift
```diff
@@ -268,8 +271,12 @@ public final class SinglePromptSecureEnclaveValet: NSObject {
     /// the keychain by the operating system due to a face, fingerprint, or password change.
     private func continuedAuthenticationKeychainQuery() throws -> [String : AnyHashable] {
         var keychainQuery = baseKeychainQuery
-        keychainQuery[kSecUseAuthenticationContext as String] = localAuthenticationContext
-        return keychainQuery
+        #if os(tvOS)
+            // tvOS doesn't support LAContext
```
NickEntin, line 275: I'm assuming omitting the authentication context won't change any behavior. Presumably this never worked on tvOS, but I don't know if the presence of the key affects behavior. cc @dfed

dfed: So I think we need to make the entire `SinglePromptSecureEnclaveValet` unavailable on tvOS.

The approach taken here will make this type behave like the `SecureEnclaveValet`, which isn't what we want.

The more I think about this, the more I think we make this a breaking change and roll a major version. We may also need a way to let folk on tvOS using this type migrate to `SecureEnclaveValet`, but given that no one has told us this wasn't working I kinda doubt anyone tried to use it that way... so maybe we don't build that until someone asks for it.

Taking the route of making `SinglePromptSecureEnclaveValet` unavailable on tvOS will also force us to solve #270, which is kinda a nice bonus.

diogot: I agree, I'll update the PR to make `SinglePromptSecureEnclaveValet` unavailable on tvOS.

dfed:
This is a good start! To get CI passing you'll likely need to update the availability checks in `Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.Swift`

diogot:
Just marking as `SinglePromptSecureEnclaveValet` as `@available(tvOS, unavailable)` is not enough to compile the project with Xcode 13.4.1, we still need the other checks as the first version.
So to make it compile with 13.4.1 and 14 I might need to keep all the checks.

diogot:
New approach, remove the classes from tvOS using `#if !os(tvOS)`.
Not it compiles fine in Xcode 13 and 14.
All tests pass with Xcode 14, but a few in `SecureEnclaveIntegrationTests` fail in Xcode 13 and I don't know why.
I may need some help to fix that.

diogot:
They are failing with `KeychainError` status: `errSecAuthFailed`

Sources/Valet/SinglePromptSecureEnclaveValet.swift
```diff
@@ -14,14 +14,14 @@
 //  limitations under the License.
 //
 
+#if !os(tvOS)
 #if canImport(LocalAuthentication)
```
dfed, line 18: Let's document what we're doing here, and also coalesce these checks into a single check (I'm pretty sure we can do this but forgive me if I'm wrong):
```suggestion
// Xcode 13 and prior incorrectly say that LocalAuthentication is available on tvOS, so we have to check both as long as Xcode 13 and prior are supported.
#if !os(tvOS) && canImport(LocalAuthentication)
```

Sources/Valet/SinglePromptSecureEnclaveValet.swift
```diff
@@ -338,3 +337,4 @@ extension SinglePromptSecureEnclaveValet {
 }
 
 #endif
+#endif
```
dfed, line 340: If you take the above suggestion, you'll need to also remove this `#endif`
```suggestion
```

Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
```diff
@@ -18,13 +18,13 @@ import Foundation
 @testable import Valet
 import XCTest
 
+#if !os(tvOS)
```
dfed, line 21:
```suggestion
// Xcode 13 and prior incorrectly say that LocalAuthentication is available on tvOS, so we have to check both as long as Xcode 13 and prior are supported.
#if !os(tvOS) && canImport(LocalAuthentication)
```

Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
```diff
@@ -18,13 +18,13 @@ import Foundation
...
 #if canImport(LocalAuthentication)
 
 class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
 {
     static let identifier = Identifier(nonEmpty: "valet_testing")!
 
-    @available(tvOS 11.0, *)
```
dfed, line 27: so good to be able to get rid of these!

Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
```diff
@@ -35,9 +35,6 @@ class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
     {
         super.setUp()
 
-        guard #available(tvOS 11.0, *) else {
-            return
-        }
```
dfed, line 40: "Fun" fact: Marking a test as `@available` does not prevent that test from being run on unavailable OS versions 😬, hence the need for all of these `guard`s

Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
```diff
@@ -269,3 +236,4 @@ class SinglePromptSecureEnclaveIntegrationTests: XCTestCase
 }
 
 #endif
+#endif
```
dfed, line 239: again suggesting this to pair with the above suggestion:
```suggestion
```

Tests/ValetObjectiveCBridgeTests/VALSinglePromptSecureEnclaveValetTests.m
```diff
@@ -16,6 +16,10 @@
 #import <Valet/Valet.h>
 #import <XCTest/XCTest.h>
 
+#if TARGET_OS_TV
+
+#else
```
dfed, line 21: Would this work?
```suggestion
#if !TARGET_OS_TV
```

Tests/ValetObjectiveCBridgeTests/VALSinglePromptSecureEnclaveValetTests.m
```diff
@@ -66,134 +70,106 @@ - (NSString *)sharedAppGroupIdentifier;
...
 - (void)test_valetWithIdentifier_accessControl_returnsCorrectValet_VALSecureEnclaveAccessControlDevicePasscode;
 {
-    if (@available(tvOS 11.0, *)) {
-        VALSinglePromptSecureEnclaveValet *const valet = [VALSinglePromptSecureEnclaveValet valetWithIdentifier:self.identifier accessControl:VALSecureEnclaveAccessControlDevicePasscode];
-        XCTAssertEqual(valet.accessControl, VALSecureEnclaveAccessControlDevicePasscode);
-        XCTAssertEqual([valet class], [VALSinglePromptSecureEnclaveValet class]);
-    }
+    VALSinglePromptSecureEnclaveValet *const valet = [VALSinglePromptSecureEnclaveValet valetWithIdentifier:self.identifier accessControl:VALSecureEnclaveAccessControlDevicePasscode];
+    XCTAssertEqual(valet.accessControl, VALSecureEnclaveAccessControlDevicePasscode);
+    XCTAssertEqual([valet class], [VALSinglePromptSecureEnclaveValet class]);
```
dfed, line 75: Holy moly this is nicer!

Tests/ValetTests/SinglePromptSecureEnclaveTests.swift
```diff
@@ -14,14 +14,14 @@
 //  limitations under the License.
 //
 
+#if !os(tvOS)
 #if canImport(LocalAuthentication)
```
dfed, line 18:
```suggestion
// Xcode 13 and prior incorrectly say that LocalAuthentication is available on tvOS, so we have to check both as long as Xcode 13 and prior are supported.
#if !os(tvOS) && canImport(LocalAuthentication)
```

Tests/ValetTests/SinglePromptSecureEnclaveTests.swift
```diff
@@ -58,3 +58,4 @@ class SinglePromptSecureEnclaveTests: XCTestCase
 }
 
 #endif
+#endif
```
dfed, line 61:
```suggestion
```

dfed:
> New approach, remove the classes from tvOS using `#if !os(tvOS)`.
> Not it compiles fine in Xcode 13 and 14.
> All tests pass with Xcode 14, but a few in `SecureEnclaveIntegrationTests` fail in Xcode 13 and I don't know why.
> I may need some help to fix that.

Which tests are failing for you? Are these tests failing for you on `master` as well, or is the failure new to your branch? Getting the test setup right locally can be relatively tricky given the code-signing requirements. Are you running these tests on a simulator, iOS device, or Mac? What OS version is the test target running on?

Good news is CI is passing, but there are some tests we don't run in CI because CI can't properly test some keychain interactions without some manual intervention 😬

dfed (approved):
This is looking really good! Some suggestions below. Let's figure out these test failures you are seeing locally, but I think we're darned close.

Tests/ValetIntegrationTests/SinglePromptSecureEnclaveIntegrationTests.swift
diogot: I figure that out, hehe

Tests/ValetObjectiveCBridgeTests/VALSinglePromptSecureEnclaveValetTests.m
diogot: It does!

diogot:
The tests are also failing in `master`, tvOS and iOS 15.5 simulators :

- `test_secureEnclaveValetsWithEqualConfiguration_canAccessSameData`
- `test_secureEnclaveValetsWithDifferingAccessControl_canNotAccessSameData`
- `test_migrateObjectsFromValet_migratesSuccessfullyToSecureEnclave`
- `test_migrateObjectsFromValet_migratesSuccessfullyAfterCanAccessKeychainCalls`

dfed:
> The tests are also failing in `master`, tvOS and iOS 15.5 simulators :
> 
> * `test_secureEnclaveValetsWithEqualConfiguration_canAccessSameData`
> * `test_secureEnclaveValetsWithDifferingAccessControl_canNotAccessSameData`
> * `test_migrateObjectsFromValet_migratesSuccessfullyToSecureEnclave`
> * `test_migrateObjectsFromValet_migratesSuccessfullyAfterCanAccessKeychainCalls`

That makes sense. I think this method needs to be updated to include later OS versions:

https://github.com/square/Valet/blob/53f25b8156e2c7678abe081a3cdf8be8cecaf5fe/Tests/ValetIntegrationTests/ValetIntegrationTests.swift#L39-L52

Note that these tests are failing on my iOS 15 simulator as well. The good news is it works on device. I'll put up a PR to fix this. In the interim, I think this PR is good to merge!

diogot:
Thanks!!

PR #286, opened 2022-09-02
Get CI running on macOS 11 and 12

dfed:
macOS 10.15 [is no longer supported by Github Actions](https://github.com/actions/runner-images/blob/main/images/macos/macos-10.15-Readme.md). So instead let's run CI on macOS 11 and macOS 12.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_12.4.app/Contents/Developer
+        run: xcversion select 12.4
       - name: Prepare Simulator Runtimes
         run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
       - name: Build and Test Framework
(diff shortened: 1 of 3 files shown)


Scripts/build.swift
```diff
@@ -266,7 +224,7 @@ guard let task = Task(rawValue: rawTask) else {
 }
 
 if task.shouldGenerateXcodeProject {
-    try execute(commandPath: "/usr/bin/swift", arguments: ["package", "generate-xcodeproj", "--output=generated/"])
+    try execute(commandPath: "/usr/bin/xcrun", arguments: ["/usr/bin/swift", "package", "generate-xcodeproj", "--output=generated/"])
```
dfed, line 227: This change is orthogonal, but using `xcrun` means we'll use the `swift` version for the selected Xcode rather than the `swift` version installed on the machine globally.

Scripts/build.swift
```diff
@@ -21,126 +21,91 @@ enum TaskError: Error {
 }
 
 enum Platform: String, CustomStringConvertible {
-    case iOS_10
-    case iOS_11
-    case iOS_12
```
dfed, line 26: I deleted platforms we aren't testing in CI to clean things up

.github/workflows/ci.yml
```diff
@@ -24,49 +23,71 @@ jobs:
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_12.4.app/Contents/Developer
+        run: xcversion select 12.4
```
dfed, line 26: This change is effectively just syntactic sugar, but it's easier to maintain.

fdiaz (approved):

NickEntin (approved):

dfed:
I've updated our repo settings to make these new CI checks all required. Merging!

PR #291, opened 2022-12-14
Add more README badges

dfed:
As title

### DIFF
--- a/Scripts/build.swift
+++ b/Scripts/build.swift
@@ -60,7 +60,7 @@ enum Platform: String, CustomStringConvertible {
         case .watchOS_7:
             return "OS=7.2,name=Apple Watch Series 6 - 44mm"
         case .watchOS_8:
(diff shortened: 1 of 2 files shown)


Scripts/build.swift
```diff
@@ -60,7 +60,7 @@ enum Platform: String, CustomStringConvertible {
         case .watchOS_7:
             return "OS=7.2,name=Apple Watch Series 6 - 44mm"
         case .watchOS_8:
-            return "OS=8.5,name=Apple Watch Series 6 - 44mm"
+            return "OS=8.5,name=Apple Watch Series 6 (44mm)"
```
NickEntin, line 63: Whyyyyyyy

NickEntin (approved):

Scripts/build.swift
dfed: Not sure. I've had to make this change on multiple repos. Seems something changed on these CI machines recently 😬

PR #297, opened 2023-05-12
Add support for multiple Valets within the same access group

efirestone:
Previously, the shared access group identifier was used for both the access group itself, as well as the uniqueness identifier for the given Valet. This adds an optional additional `identifier` parameter that can be specified when creating a shared access group Valet. The identifier adds an additional element of uniqueness, so two Valets with the same shared access group can exist, and their data will not overlap, so long as they have different identifiers.
    
The default for this value is `nil`, which keeps the existing behavior for full backward compatibility.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,3 @@
-source 'https://rubygems.org' do
-  gem 'cocoapods', '~> 1.11.0'
-end
+source "https://rubygems.org"
+
+gem 'cocoapods', '~> 1.11.0'
--- a/README.md
+++ b/README.md
@@ -169,6 +169,8 @@ VALValet *const mySharedValet = [VALValet sharedGroupValetWithGroupPrefix:@"grou
 
 This instance can be used to store and retrieve data securely across any app written by the same developer that has `group.Druidia` set as a value for the `com.apple.security.application-groups` key in the app’s `Entitlements`. This Valet is accessible when the device is unlocked. Note that `myValet` and `mySharedValet` cannot read or modify one another’s values because the two Valets were created with different initializers. All Valet types can share secrets across applications written by the same developer by using the `sharedGroupValet` initializer. Note that on macOS, the `groupPrefix` [must be the App ID prefix](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_application-groups#discussion).
 
+As with Valets, shared iCloud Valets can be created with an additional identifier, allowing multiple independently sandboxed keychains to exist within the same shared group.
+
 ### Sharing Secrets Across Devices with iCloud
 
 ```swift
@@ -180,6 +182,8 @@ VALValet *const myCloudValet = [VALValet iCloudValetWithIdentifier:@"Druidia" ac
 ```
 
 This instance can be used to store and retrieve data that can be retrieved by this app on other devices logged into the same iCloud account with iCloud Keychain enabled. If iCloud Keychain is not enabled on this device, secrets can still be read and written, but will not sync to other devices. Note that  `myCloudValet` can not read or modify values in either `myValet` or `mySharedValet` because `myCloudValet` was created a different initializer.
+
+Shared iCloud Valets can be created with an additional identifier, allowing multiple independently sandboxed keychains to exist within the same iCloud shared group.
 
 ### Protecting Secrets with Face ID, Touch ID, or device Passcode
 
--- a/Sources/Valet/Valet.swift
+++ b/Sources/Valet/Valet.swift
@@ -40,19 +40,21 @@ public final class Valet: NSObject {
     }
 
     /// - Parameters:
-    ///   - identifier: The identifier for the Valet's shared access group. Must correspond with the value for keychain-access-groups in your Entitlements file.
+    ///   - groupIdentifier: The identifier for the Valet's shared access group. Must correspond with the value for keychain-access-groups in your Entitlements file.
+    ///   - identifier: An optional additional uniqueness identifier. Using this identifier allows for the creation of separate, sandboxed Valets within the same shared access group.
     ///   - accessibility: The desired accessibility for the Valet.
     /// - Returns: A Valet that reads/writes keychain elements that can be shared across applications written by the same development team.
-    public class func sharedGroupValet(with identifier: SharedGroupIdentifier, accessibility: Accessibility) -> Valet {
-        findOrCreate(identifier, configuration: .valet(accessibility))
+    public class func sharedGroupValet(
+        with groupIdentifier: SharedGroupIdentifier, identifier: Identifier? = nil, accessibility: Accessibility) -> Valet {
+            findOrCreate(groupIdentifier, identifier: identifier, configuration: .valet(accessibility))
     }
 
     /// - Parameters:
     ///   - identifier: The identifier for the Valet's shared access group. Must correspond with the value for keychain-access-groups in your Entitlements file.
     ///   - accessibility: The desired accessibility for the Valet.
     /// - Returns: A Valet (synchronized with iCloud) that reads/writes keychain elements that can be shared across applications written by the same development team.
-    public class func iCloudSharedGroupValet(with identifier: SharedGroupIdentifier, accessibility: CloudAccessibility) -> Valet {
-        findOrCreate(identifier, configuration: .iCloud(accessibility))
+    public class func iCloudSharedGroupValet(with groupIdentifier: SharedGroupIdentifier, identifier: Identifier? = nil, accessibility: CloudAccessibility) -> Valet {
+        findOrCreate(groupIdentifier, identifier: identifier, configuration: .iCloud(accessibility))
     }
 
     #if os(macOS)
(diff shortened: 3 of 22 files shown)


Gemfile
```diff
@@ -1,3 +1,5 @@
+source "https://rubygems.org"
```
dfed, line 1: can we remove the `source 'https://rubygems.org' do` below? Seems duplicative to have both

Sources/Valet/Valet.swift
```diff
@@ -40,19 +40,21 @@ public final class Valet: NSObject {
     }
 
     /// - Parameters:
-    ///   - identifier: The identifier for the Valet's shared access group. Must correspond with the value for keychain-access-groups in your Entitlements file.
+    ///   - groupIdentifier: The identifier for the Valet's shared access group. Must correspond with the value for keychain-access-groups in your Entitlements file.
     ///   - accessibility: The desired accessibility for the Valet.
+    ///   - identifier: An optional additional uniqueness identifier. Using this identifier allows for the creation of multiple Valets within the same shared access group.
```
dfed, line 45: possible clarity improvement:
```suggestion
    ///   - identifier: An optional additional uniqueness identifier. Using this identifier allows for the creation of separate, sandboxed Valets within the same shared access group.
```

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -317,6 +318,19 @@ class ValetIntegrationTests: XCTestCase
         }
     }
 
+    func test_stringForKey_withDifferingIdentifierInSameAccessGroup_throwsItemNotFound() throws
```
dfed, line 321: 🎉

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -650,6 +677,22 @@ class ValetIntegrationTests: XCTestCase
...
+        try valet2.setString(passcode, forKey: key)
+
+        try valet2.removeObject(forKey: key)
+
+        XCTAssertEqual(passcode, try valet1.string(forKey: key))
+        XCTAssertThrowsError(try valet2.string(forKey: key)) { error in
+            XCTAssertEqual(error as? KeychainError, .itemNotFound)
+        }
+    }
+
```
dfed, line 695: It might be nice to add a test that migrating between two different valets with different identifiers works as expected. (I know the test matrix is big + annoying but the more tests we have the more confident we are that this stuff actually works)

dfed (commented):
Overall this seems reasonable to me. We're taking on a bit more complexity, but it's not too shabby. I'd love to better understand the use case for this, and why we're only adding this functionality to shared access Valets: is the use case specific to shared access Valets or is your current need currently scoped to shared access Valets?

Also: thoughts on updating the README to document this new API?

Gemfile
dfed: yeah we should be able to remove the `do/end` block below. [Example](https://github.com/dfed/swift-async-queue/blob/a232ddbfb776a3d3e380dce30e7d5f617cffcf82/Gemfile#L3-L5).

efirestone:
> Is the use case specific to shared access Valets or is your current need currently scoped to shared access Valets?
Our use case is only for shared group access Valets. The same thing could easily be extended to iCloud and other shared Valets as well, but I figured I'd start here to scope it down. Happy to add the other cases in this PR or a follow up too though.
 
> Also: thoughts on updating the README to document this new API?
Yup, good call. On it.

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
efirestone: The migration matrix doesn't seem to care about the *types* of the Valets that are involved in the migration, and more about the variations and keys and values. For example, there's nothing testing the migration of iCloud Valets to group access Valets to local Valets, etc. And I think that makes sense given that the migration uses the standard get/set/remove methods in all cases, and those are tested separately for each type.

Happy to add something here, but not entirely sure which tests of value are missing.

dfed: Eh that's fair. I was going off memory. If we don't have similar tests then we're good to go here. I would, however, like to see test coverage improve (based on the code of comment), but I'm currently in mobile so I haven't checked details. Let me know if I can help offer possible tests to up coverage.

efirestone:
> > Is the use case specific to shared access Valets or is your current need currently scoped to shared access Valets?
> 
> Our use case is only for shared group access Valets. The same thing could easily be extended to iCloud and other shared Valets as well, but I figured I'd start here to scope it down. Happy to add the other cases in this PR or a follow up too though.
> 
> > Also: thoughts on updating the README to document this new API?
> 
> Yup, good call. On it.

Updated the README and added the ability to add uniqueness identifiers to the other shared types (iCloud, SecureEnclave, SinglePromptSecureEnclave).

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
efirestone: The Codecov report says `The diff coverage is 100.00%.`, so I think all the new content is covered. It went down overall, but I'm not sure how that's calculated.

README.md
```diff
@@ -181,6 +190,8 @@ VALValet *const myCloudValet = [VALValet iCloudValetWithIdentifier:@"Druidia" ac
 
 This instance can be used to store and retrieve data that can be retrieved by this app on other devices logged into the same iCloud account with iCloud Keychain enabled. If iCloud Keychain is not enabled on this device, secrets can still be read and written, but will not sync to other devices. Note that  `myCloudValet` can not read or modify values in either `myValet` or `mySharedValet` because `myCloudValet` was created a different initializer.
 
+Similar to with shared group Valets, shared iCloud Valets can be created with an additional identifier, allowing multiple independently sandboxed keychains to exist within the same iCloud shared group.
```
dfed, line 193:
```suggestion
As with Valets, shared iCloud Valets can be created with an additional identifier, allowing multiple independently sandboxed keychains to exist within the same iCloud shared group.
```

README.md
```diff
@@ -169,6 +169,15 @@ VALValet *const mySharedValet = [VALValet sharedGroupValetWithGroupPrefix:@"grou
...
 
+It is possible to create multiple Valets within the same shared access group be specifying the same group identifier but different unique identifiers, like so:
+
+```
+let valet1 = Valet.sharedGroupValet(with: groupID, identifier: Identifier(nonEmpty: "valet1")!, accessibility: accessibility)
+let valet2 = Valet.sharedGroupValet(with: groupID, identifier: Identifier(nonEmpty: "valet2")!, accessibility: accessibility)
+```
+
+These two Valets cannot access each others data, but they are both accessible from all applications or extensions which have access to the given `groupID`. The `identifier` (but not the `groupIdentifier`) is optional on shared Valets. All Valets created within the same group using the default `identifier` can access the same data.
+
```
dfed, line 180: let's move this blurb up to the `Valet` section and then replace this with:
```suggestion
As with Valets, shared iCloud Valets can be created with an additional identifier, allowing multiple independently sandboxed keychains to exist within the same shared group.

```

dfed (approved):
LGTM with a few changes to the README.

README.md
dfed: @efirestone we still want this blurb but we want it further up in the README 😉

efirestone:
@dfed can you merge this? I don't have permissions 🙏

NickEntin (approved):

README.md
efirestone: I'm not sure I'm following what you're suggesting (and now the review recommendation can't be committed). Are you saying put this blurb in the `Valet` section:
```
As with Valets, shared iCloud Valets can be created with an additional identifier, allowing multiple independently sandboxed keychains to exist within the same shared group.
```
or to leave that blurb in the "Sharing Secrets" section, and then put the original text up in the Valet section?

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
efirestone: Done!

README.md
dfed: Ah! Okay I see where I failed to read here. It's the _additional_ in `additional identifier` that threw me.

The _additional_ is referring to iCloud Valets, whereas Valets have a primary identifier.

I think we should just remove the "As with Valets" preamble and this'll make more sense to me.

dfed (approved):

efirestone:
"As with Valets" prefix removed in the README, so should be good to go.

The iOS 13 + iOS 14 builds aren't finishing anymore, I'm guessing due to a GitHub change. It looks like `master` is having the [same problem now](https://github.com/square/Valet/pull/299). I'm investigating some in that PR, but so far I'm kind of stuck.

dfed:
> "As with Valets" prefix removed in the README, so should be good to go.
> 
> 
> 
> The iOS 13 + iOS 14 builds aren't finishing anymore, I'm guessing due to a GitHub change. It looks like `master` is having the [same problem now](https://github.com/square/Valet/pull/299). I'm investigating some in that PR, but so far I'm kind of stuck.

Thanks for addressing!

Re Xcode 12.4: can you file an issue on https://github.com/actions/runner-images? I've had success there in the past when there was a runner configuration issue.

dfed:
Just filed an issue for what we're seeing with CI: https://github.com/actions/runner-images/issues/7687

dfed:
Just merged latest into this branch. Should turn green, at which point merge at will @efirestone. Let me know if you want help publishing the feature version update.

dfed:
Just released this as 4.2.0

PR #300, opened 2023-06-05
Modernize CI

jshier:
This PR starts modernizing the project's CI integration, starting with updating the platform support.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,7 +107,7 @@ jobs:
       - name: Install Carthage
         run: brew outdated carthage || brew upgrade carthage
       - name: Build Framework
-        run: carthage build --verbose --no-skip-current
+        run: carthage build --verbose --no-skip-current --use-xcframework
   spm-11:
     name: SPM Build macOS 11
     runs-on: macOS-11
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
       matrix:
         platforms: [
           'iOS_14,tvOS_14,watchOS_7',
+           # TODO: add iOS_13 back once https://github.com/actions/runner-images/issues/7687 is resolved
           'tvOS_13,watchOS_6',
         ]
       fail-fast: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
           'iOS_13,tvOS_13,watchOS_6',
         ]
       fail-fast: false
+    timeout-minutes: 30
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
@@ -44,6 +45,7 @@ jobs:
           'iOS_15,tvOS_15,watchOS_8',
         ]
       fail-fast: false
+    timeout-minutes: 30
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
@@ -65,13 +67,14 @@ jobs:
           'iOS_16,tvOS_16,watchOS_9',
         ]
       fail-fast: false
+    timeout-minutes: 30
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_14.3.1.app/Contents/Developer
+        run: sudo xcode-select --switch /Applications/Xcode_14.3.app/Contents/Developer
       - name: Build and Test Framework
         run: Scripts/build.swift ${{ matrix.platforms }} xcode
       - name: Upload Coverage Reports
@@ -79,26 +82,28 @@ jobs:
         run: Scripts/upload-coverage-reports.sh ${{ matrix.platforms }}
   pod-lint:
     name: Pod Lint
-    runs-on: macOS-11
+    runs-on: macOS-13
+    timeout-minutes: 30
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_11.7.app/Contents/Developer
+        run: sudo xcode-select --switch /Applications/Xcode_14.3.app/Contents/Developer
       - name: Lint Podspec
         run: bundle exec pod lib lint --verbose --fail-fast --swift-version=5.0
   carthage:
     name: Carthage
-    runs-on: macOS-11
+    runs-on: macOS-13
+    timeout-minutes: 30
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_11.7.app/Contents/Developer
+        run: sudo xcode-select --switch /Applications/Xcode_14.3.app/Contents/Developer
       - name: Install Carthage
         run: brew outdated carthage || brew upgrade carthage
       - name: Build Framework
@@ -106,6 +111,7 @@ jobs:
   spm-11:
     name: SPM Build macOS 11
     runs-on: macOS-11
+    timeout-minutes: 30
     strategy:
       matrix:
         platforms: [
@@ -146,6 +152,7 @@ jobs:
           'macOS_12',
         ]
       fail-fast: false
+    timeout-minutes: 30
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
@@ -167,13 +174,14 @@ jobs:
           'macOS_13',
         ]
       fail-fast: false
+    timeout-minutes: 30
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_14.3.1.app/Contents/Developer
+        run: sudo xcode-select --switch /Applications/Xcode_14.3.app/Contents/Developer
       - name: Prepare Simulator Runtimes
         run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
       - name: Build Framework
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         platforms: [
           'iOS_14,tvOS_14,watchOS_7',
-          'iOS_13,tvOS_13,watchOS_6',
+          'tvOS_13,watchOS_6',
         ]
       fail-fast: false
     timeout-minutes: 30
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,10 @@ on:
     - master
   pull_request:
 
+concurrency:
+  group: ${{ github.ref_name }}
+  cancel-in-progress: true
+
 jobs:
   xcode-build-12:
     name: Xcode 12 Build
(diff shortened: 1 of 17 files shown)


dfed (approved):
Thank you for the update!

.github/workflows/ci.yml
```diff
@@ -123,7 +144,7 @@ jobs:
       fail-fast: false
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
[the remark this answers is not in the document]
jshier: Sure. I wanted to get CI runs going, then I'll iterate to add and update all platforms and combos.

dfed:
Unit tests failing due `testEnvironmentSupportsWhenPasscodeSet` not being updated to include newer OS versions.

Xcode 12 hanging is likely [the same as this issue](https://github.com/square/Valet/pull/297#issuecomment-1573088010)

jshier:
@dfed Are you okay dropping any versions of Xcode in the CI matrix? Or do you want to keep going with Xcode 11?

dfed:
> @dfed Are you okay dropping any versions of Xcode in the CI matrix? Or do you want to keep going with Xcode 11?

I'd hate to lose building for macOS 10.15 😕. That said... we are a bit long overdue for a breaking version bump. Could be nice to drop support for watchOS 2-3, tvOS 9-11, iOS 9-11, and macOS 10.11-10.13. Only issue there is we can't set our minimum Xcode version to something recent and still test older simulators in CI.

Regarding the current failing tests... seems like tests protected by `testEnvironmentIsSigned()` are failing due to an entitlement error. This is new/exciting, and may not be easy to fix 😬

jshier:
> Regarding the current failing tests... seems like tests protected by `testEnvironmentIsSigned()` are failing due to an entitlement error. This is new/exciting, and may not be easy to fix 😬

I hoped that those failures were dealt with in the existing CI infrastructure, but I guess not. I'll take a look.

jshier:
> I'd hate to lose building for macOS 10.15 😕.

I can keep it, but technically that image is deprecated by GitHub, so I don't know how much longer it'll exist at all.

dfed:
> I hoped that those failures were dealt with in the existing CI infrastructure, but I guess not. I'll take a look.

Thank you! In the past we've avoided running unit tests that required entitlements in CI. We may need to take the same approach here, at which point we'd rename `isTestEnvironmentSigned()` to something like `testEnvironmentSupportsEntitlements()` or some such and update the check. Valet is the kind of repo that, unfortunately, really requires local testing during development.

> I can keep it, but technically that image is deprecated by GitHub, so I don't know how much longer it'll exist at all.

Oh! That's information I didn't have. If those images are indeed deprecated and slated for removal then yeah we can remove them. cc @efirestone

jshier:
What's the recommended way to test locally with the proper entitlements?

dfed:
> What's the recommended way to test locally with the proper entitlements?

Our [Contributing.md file](https://github.com/square/Valet/blob/master/Contributing.md#testing-changes-on-macos) outlines how to test with local entitlements on macOS – if iOS now requires testing on device as of Xcode 14, you'll need to go through similar steps on iOS. Though note that the iOS `Host Application` is already set.

jshier:
@dfed I updated the condition and the tests pass locally without entitlements, so I'd like to try it here before I full rename and add all the additional CI runs.

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -24,6 +24,12 @@ import XCTest
 /// - The Valet Mac Tests target is left without a host app on master. Mac test host app signing requires CI to have the Developer team credentials down in keychain, which we can't easily accomplish.
 /// - Note: In order to test changes locally, set the Valet Mac Tests host to Valet macOS Test Host App, delete all VAL_* keychain items in your keychain via Keychain Access.app, and run Mac tests.
 func testEnvironmentIsSigned() -> Bool {
+    #if targetEnvironment(simulator)
+    if #available(iOS 16, tvOS 16, macOS 13, watchOS 9, *) {
```
dfed, line 28:
```suggestion
    #if targetEnvironment(simulator)
    // Xcode 14's simulator environment is signed, but requires entitlements we can't provide in CI.
    if #available(iOS 16, tvOS 16, macOS 13, watchOS 9, *) {
```

Also, can you modify the `/// - Returns: `true` when the test environment is signed.` line above to read `/// - Returns: `true` when the test environment is signed and entitlements are not required by the current environment.`

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
```diff
@@ -42,8 +48,12 @@ func testEnvironmentSupportsWhenPasscodeSet() -> Bool {
         || simulatorVersionInfo.contains("tvOS 13")
         || simulatorVersionInfo.contains("iOS 15")
         || simulatorVersionInfo.contains("tvOS 15")
+        || simulatorVersionInfo.contains("iOS 16")
+        || simulatorVersionInfo.contains("tvOS 16")
+        || simulatorVersionInfo.contains("iOS 17")
+        || simulatorVersionInfo.contains("tvOS 17")
     {
-        // iOS 13, tvOS 13, iOS 15, and tvOS 15 simulators fail to store items in a Valet that has a
+        // iOS 13, tvOS 13, iOS 15, tvOS 15, and later simulators fail to store items in a Valet that has a
```
dfed, line 56: still funny to me that iOS/tvOS 14 didn't have this issue.

.github/workflows/ci.yml
```diff
@@ -6,6 +6,10 @@ on:
     - master
   pull_request:
 
+concurrency:
+  group: ${{ github.ref_name }}
+  cancel-in-progress: true
```
dfed, line 11: What benefit do we get from this? I thought we were already running in parallel.

I also wouldn't recommend cancelling in progress jobs when one fails. Given how wide our test matrix is it's helpful to know many entries a particular problem is scoped to.

dfed (commented):
Really appreciate your progress here.

.github/workflows/ci.yml
jshier: Last I tried, cancellation wasn't automatic, but GitHub will happily show the newest run in the UI, even while old runs are still going. This ensures that doesn't happen by assigning a PR-unique identifier and cancelling the previous jobs. 

This doesn't affect cancellation on failure.

Tests/ValetIntegrationTests/ValetIntegrationTests.swift
jshier: Updated.

.github/workflows/ci.yml
dfed: Ah – this is about cancelling prior runs **when new commits are pushed**. I don't have strong feelings here: this addition makes sense to me 🙂

jshier:
Ah, looks like Xcode 14.3 removed `generate-xcodeproj` (it'd been deprecated for a while), so a new way to run those SPM tests is needed.

dfed:
> I moved the Carthage and CocoaPods runners to use Xcode 14.3, do you want some dedicated Xcode 11 tests instead?

What you've done here makes sense to me. However, it seems our supporting absurdly old iOS versions is causing our Carthage build to fail with modern Xcode. Oof.

I am leaning towards ripping off the bandaid and rolling a major version bump for this. Let's drop support for anything prior to iOS 11, tvOS 11, watchOS 4, and macOS 10.13 and update the README to reflect that.

I'm also seeing an error that `swift package generate-xcodeproj` was removed in Xcode 14... which... ugh. To fix that, we're going to:
1. Make `shouldGenerateXcodeProject` return `false` for anything running on Xcode 14+
2. Remove the `"-project", task.project,` arguments from the `xcodeBuildArguments` list on Xcode 14+

I *think* that'll do it [given a similar script I have in another repo](https://github.com/dfed/swift-async-queue/blob/main/Scripts/build.swift).

jshier:
> I'm also seeing an error that `swift package generate-xcodeproj` was removed in Xcode 14... which... ugh. To fix that, we're going to:
> 
> 1. Make `shouldGenerateXcodeProject` return `false` for anything running on Xcode 14+
> 2. Remove the `"-project", task.project,` arguments from the `xcodeBuildArguments` list on Xcode 14+

One other wrinkle I recall is that `xcodebuild` doesn't support building Xcode projects and SPM projects out of the same directory. So to build the package using `xcodebuild` we'd have to first delete the project file, otherwise the schemes can't be properly looked up.

jshier:
> welp. I'd be fine moving the xcodeproj/xcworkspace or Package.swift into a new directory. I think I'd prefer moving the `xcodeproj/xcworkspace`?

I don't think it can work any other way, given Xcode / SPM doesn't support nested package files AFAIK.

Sources/Valet/SinglePromptSecureEnclaveValet.swift
```diff
@@ -15,6 +15,7 @@
 //
 
 // Xcode 13 and prior incorrectly say that LocalAuthentication is available on tvOS, so we have to check both as long as Xcode 13 and prior are supported.
+// Xcode 14 moved the LAContext availability to watchOS 3, so only that version is explicitly annotated.
```
dfed, line 18: Thank you for continuing to document this absurdity.

Scripts/build.swift
```diff
@@ -95,22 +127,40 @@ enum Platform: String, CustomStringConvertible {
...
-             .watchOS_8:
+             .watchOS_8,
+             .watchOS_9,
+             .watchOS_10:
             // watchOS does not support unit testing (yet?).
             return false
         }
     }
 
+    var requiresUpdatedSPMIntegration: Bool {
```
dfed, line 154: Would love a comment here. Otherwise maybe we could rename to `supportsGeneratedXcodeproj` and invert the check? As-is it's not clear what is "updated" about the integration.

Scripts/build.swift
```diff
@@ -213,33 +283,54 @@ enum Task: String, CustomStringConvertible {
...
+    guard let platform = Platform(rawValue: rawPlatform) else {
+        print("Received unknown platform type \(rawPlatform)")
+        exit(0)
+    }
+
+    return platform
 }
 
-if task.shouldGenerateXcodeProject {
+if task == .spm && platforms.map({ task.shouldGenerateXcodeProject(for: $0) }).contains(true) {
```
dfed, line 305: Could we extract this into a computed property? Helps make the code more approachable 🙂

Scripts/build.swift
```diff
@@ -213,33 +283,54 @@ enum Task: String, CustomStringConvertible {
...
     try execute(commandPath: "/usr/bin/xcrun", arguments: ["/usr/bin/swift", "package", "generate-xcodeproj", "--output=generated/"])
 }
 
+for platform in platforms {
+    var deletedXcodeproj = false
+    var xcodeBuildArguments: [String] = []
+    if task == .spm && platform.requiresUpdatedSPMIntegration {
+        do {
+            print("Deleting Valet.xcodeproj, any uncommitted changes will be lost.")
```
dfed, line 314:
```suggestion
            // Swift will error if an xcodeproj and Package.swift file are both present in this directory. Delete the xcodeproj to prevent this. We'll restore it later.
            print("Deleting Valet.xcodeproj, any uncommitted changes will be lost.")
```

Sources/LegacyValet/VALLegacySecureEnclaveValet.m
```diff
@@ -273,6 +275,7 @@ - (BOOL)removeAllObjects;
 {
     VALCheckCondition(NO, NO, @"%s is not supported on %@", __PRETTY_FUNCTION__, NSStringFromClass([self class]));
 }
+#pragma clang diagnostic pop
```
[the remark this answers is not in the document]
dfed: Yeah agreed we're good as-is

dfed (commented):
Looking good!

Scripts/build.swift
jshier: You mean a local variable? I didn't make it a computed property because it's drawing from both the task and platforms.

dfed: That's what I get for reviewing on mobile 😅. Forgot this was top level. Yeah a local variable or a method would be great.

Scripts/build.swift
jshier: I added some different comments which should resolve this.

Scripts/build.swift
jshier: Done.

Scripts/build.swift
jshier: Added.

.github/workflows/ci.yml
```diff
@@ -14,16 +18,17 @@ jobs:
       matrix:
         platforms: [
           'iOS_14,tvOS_14,watchOS_7',
-          'iOS_13,tvOS_13,watchOS_6',
+          'tvOS_13,watchOS_6',
```
dfed, line 21:
```suggestion
           # TODO: add iOS_13 back once https://github.com/actions/runner-images/issues/7687 is resolved
          'tvOS_13,watchOS_6',
```

dfed, line 21:
```suggestion
           # TODO: add iOS_13 back once https://github.com/actions/runner-images/issues/7687 is resolved
          'tvOS_13,watchOS_6',
```

dfed:
Thanks for the work here @jshier! Merging 😄

PR #301, opened 2023-07-04
Add iOS 13 back to CI

dfed:
As title. Per https://github.com/actions/runner-images/issues/7687#issuecomment-1620396057 it's time to add this back.


NickEntin (approved):
🤷‍♂️  Cool.

PR #308, opened 2024-04-22
Make core types Sendable and @unchecked Sendable

dfed:
We've been thread-safe since v1. We just hadn't told the compiler this fact yet.

This PR won't compile on Xcode 12.4 or prior, and we currently support back to Xcode 11. We therefore have two options:

1. Drop support for these long-useless Xcode builds in a feature version. This is somewhat reasonable since you haven't been able to ship to the App Store with these versions for years. We also haven't run these old versions of Xcode in CI in many years.
1. Roll a breaking change. I'd prefer to do that with more thought though: if we want a breaking change it'll be a bit before this can land + be released.

I'd be curious to hear from @pwesten and friends how they'd like to approach this before I go and update the README on this PR.

### DIFF
--- a/Sources/Valet/SecureEnclaveValet.swift
+++ b/Sources/Valet/SecureEnclaveValet.swift
@@ -19,7 +19,7 @@ import Foundation
 
 /// Reads and writes keychain elements that are stored on the Secure Enclave using Accessibility attribute `.whenPasscodeSetThisDeviceOnly`. Accessing these keychain elements will require the user to confirm their presence via Touch ID, Face ID, or passcode entry. If no passcode is set on the device, accessing the keychain via a `SecureEnclaveValet` will fail. Data is removed from the Secure Enclave when the user removes a passcode from the device.
 @objc(VALSecureEnclaveValet)
-public final class SecureEnclaveValet: NSObject {
+public final class SecureEnclaveValet: NSObject, Sendable {
     
     // MARK: Public Class Methods
 
@@ -92,7 +92,6 @@ public final class SecureEnclaveValet: NSObject {
         self.identifier = identifier
         self.service = service
         self.accessControl = accessControl
-        baseKeychainQuery = service.generateBaseQuery()
     }
     
     // MARK: Hashable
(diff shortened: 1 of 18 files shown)


Sources/Valet/SinglePromptSecureEnclaveValet.swift
```diff
@@ -25,8 +25,8 @@ import Foundation
 /// Reads and writes keychain elements that are stored on the Secure Enclave using Accessibility attribute `.whenPasscodeSetThisDeviceOnly`. The first access of these keychain elements will require the user to confirm their presence via Touch ID, Face ID, or passcode entry. If no passcode is set on the device, accessing the keychain via a `SinglePromptSecureEnclaveValet` will fail. Data is removed from the Secure Enclave when the user removes a passcode from the device.
 @objc(VALSinglePromptSecureEnclaveValet)
 @available(watchOS 3, *)
-public final class SinglePromptSecureEnclaveValet: NSObject {
-    
+public final class SinglePromptSecureEnclaveValet: NSObject, @unchecked Sendable {
```
dfed, line 28: The localization context is a `var` on the type. It's only mutated within a lock, so we are indeed sendable. But the compiler won't be convinced of that... so we use unchecked.

Sources/Valet/SecureEnclaveValet.swift
```diff
@@ -232,7 +231,9 @@ public final class SecureEnclaveValet: NSObject {
     // MARK: Private Properties
 
     private let lock = NSLock()
-    private let baseKeychainQuery: [String : AnyHashable]
+    private var baseKeychainQuery: [String : AnyHashable] {
+        return service.generateBaseQuery()
+    }
```
dfed, line 236: I don't love this, but `AnyHashable` isn't `Sendable` and `service.generateBaseQuery()` is fast, so in the tug-of-war between "have the compiler check the thing I know is safe" and "we know it's safe so opt out of checking to do something faster" I'm leaning towards the former.

NickEntin (approved):

NickEntin:
> This PR won't compile on Xcode 12.4 or prior, and we currently support back to Xcode 11. We therefore have two options:
> 
> 1. Drop support for these long-useless Xcode builds in a feature version. This is somewhat reasonable since you haven't been able to ship to the App Store with these versions for years. We also haven't run these old versions of Xcode in CI in many years.
> 2. Roll a breaking change. I'd prefer to do that with more thought though: if we want a breaking change it'll be a bit before this can land + be released.

I generally lean towards calling dropped Xcode support a breaking change, but the "haven't been able to ship to the App Store with these versions for years" is an interesting mitigating factor. I'm okay with dropping that as a minor version bump. 👍

PR #310, opened 2024-04-23
Get CI running Xcode 15

dfed:
It builds. :shipit: 

Also updated to use macOS 14 and Xcode 15 when available, and also upgraded to `checkout@v4`.

Edit: No, it doesn't build. But #313 fixes that.


.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     timeout-minutes: 30
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
dfed, line 27: just updating while we're here

dfed:
Merging since this is just CI and per #313 this change doesn't really make a big difference 😅

PR #311, opened 2024-04-23
Get watchOS tests running in CI

### LINKED ISSUE #128: Run Watch tests in CI
Since watchOS does not have an XCTest framework, we [created a mock of the XCTest framework](https://github.com/square/Valet/pull/127/files#diff-4faaefffc8a2c2b2f510a97d4e1f2f74R30) and [built a watch app to host & run the tests](https://github.com/square/Valet/pull/127/files#diff-2dda50437e8523a6808dbbcf3eeb8791R29).

This watch app allows us to manually verify that Valet works as expected on the watch, but it doesn't allow us to catch regressions in CI. It'd be great for us to improve this situation by running our watch app in CI.

Since the watch app exits with code 0 after successfully running tests, we're most of the way there. The crux of the remaining challenge is determining how to launch the watch app in the simulator from the command line.

I've done some digging, and it looks like we may need to rely one some of the following commands:
- `xcrun simctl list` # this command will allow us to determine which sims are on the CI machine, and allow us to find the UUIDs of phone+watch pairs.
- `xcrun simctl boot <uuid>` # this command will boot the sims with the uuids we've found above. We'll likely need to launch both the phone and watch sims.
- `xcrun simctl install <uuid> <path-to-phone-app AND/OR path-to-watch-app>` # this command should install the app on the simulator
- `xcrun simctl launch <uuid> <path-to-phone-app AND/OR path-to-watch-app>` # this command should allow us to launch the app(s) we installed above

I'm likely missing a command above, as I haven't managed to get a script working with the above. Pro-tip: you can monitor the logs emanating from your simulators by running the command `xcrun simctl spawn <uuid> log stream`. Note that the phone host app is `ValetTouchIDTest`, and the watch app is `Valet watchOS Test Host App`.

Help would be greatly appreciated, since I'm brand new to command-line simulator manipulation.

dfed:
Resolves #128.

We don't have coverage for all watchOS versions, but it's better than what we had.

A good amount of work is from #312 and #313 – it's worth reviewing my comments there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,8 +17,12 @@ jobs:
     strategy:
       matrix:
         platforms: [
-          'iOS_14,tvOS_14,watchOS_7',
-          'iOS_13,tvOS_13,watchOS_6',
+          'iOS_14',
+          'tvOS_14',
+          'watchOS_7',
+          'iOS_13',
+          'tvOS_13',
+          'watchOS_6',
         ]
       fail-fast: false
     timeout-minutes: 30
@@ -42,7 +46,9 @@ jobs:
     strategy:
       matrix:
         platforms: [
-          'iOS_15,tvOS_15,watchOS_8',
+          'iOS_15',
+          'tvOS_15',
+          'watchOS_8',
         ]
       fail-fast: false
     timeout-minutes: 30
@@ -64,8 +70,12 @@ jobs:
     strategy:
       matrix:
         platforms: [
-          'iOS_16,tvOS_16,watchOS_9',
-          'iOS_17,tvOS_17,watchOS_10',
+          'iOS_16',
+          'tvOS_16',
+          'watchOS_9',
+          'iOS_17',
+          'tvOS_17',
+          'watchOS_10',
         ]
       fail-fast: false
     timeout-minutes: 30
@@ -116,8 +126,12 @@ jobs:
     strategy:
       matrix:
         platforms: [
-          'iOS_14,tvOS_14,watchOS_7',
-          'iOS_13,tvOS_13,watchOS_6',
+          'iOS_14',
+          'tvOS_14',
+          'watchOS_7',
+          'iOS_13',
+          'tvOS_13',
+          'watchOS_6',
           'macOS_11',
         ]
       fail-fast: false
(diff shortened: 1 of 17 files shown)


Tests/ValetIntegrationTests/SecureEnclaveIntegrationTests.swift
```diff
@@ -72,7 +72,7 @@ class SecureEnclaveIntegrationTests: XCTestCase
 
     func test_secureEnclaveSharedGroupValetsWithDifferingIdentifiers_canNotAccessSameData() throws
     {
-        guard testEnvironmentIsSigned() && testEnvironmentSupportsWhenPasscodeSet() else {
+        guard testEnvironmentIsSignedOrDoesNotRequireEntitlement() && testEnvironmentSupportsWhenPasscodeSet() else {
```
dfed, line 75: this is what led me to #313

.github/workflows/ci.yml
```diff
@@ -17,8 +17,12 @@ jobs:
...
       matrix:
         platforms: [
-          'iOS_14,tvOS_14,watchOS_7',
-          'iOS_13,tvOS_13,watchOS_6',
+          'iOS_14',
+          'tvOS_14',
+          'watchOS_7',
+          'iOS_13',
+          'tvOS_13',
+          'watchOS_6',
```
dfed, line 25: split this out for faster iteration because we have the machines

README.md
```diff
@@ -239,7 +239,7 @@ Valet guarantees that reading and writing operations will succeed as long as wri
 
 ## Requirements
 
-* Xcode 12.5 or later.
+* Xcode 13.0 or later.
```
dfed, line 242: Follow-up to #308 now that CI is working.

.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "Tests/xctest-watchOS"]
-	path = Tests/XCTest-watchOS
-	url = https://github.com/dfed/XCTest-watchOS.git
```
dfed, line 3: :wave:

dfed:
This is not an API change - it only affects CI. Merging pre-review to get CI working again + adding watch coverage. Happy to address feedback post-merge.

PR #318, opened 2024-08-22
Ignore tests in code coverage

dfed:


dfed:
never mind – we're already doing this via our custom upload script!

PR #334, opened 2025-01-08
Use better broken link checker

dfed:
Looks like the link checker we updated to recently doesn't catch broken internal-to-repo links (see https://github.com/dfed/SafeDI/pull/143). This PR updates to one that does.


luispadron (approved):
nice!

Issue #339, opened 2025-03-12
Remove "internal" access modifiers from code


richardtop:
Since "internal" access modifier is the default one, it can be safely removed. Or are there any reasons to actually explicitly state that access modifier for methods/classes?
e.g.: https://github.com/square/Valet/blob/main/Sources/Valet/Internal/Keychain.swift#L20

if this is OK, I can make a PR, or feel free to close the issue if you'd like to keep it this way.

dfed:
Indeed the modifier is not necessary. Explicit access control was a style I played with back when we rewrote Valet in Swift. Happy to remove it. Feel free to submit a PR.

richardtop:
@dfed could you please "release" a new version with this change? Thanks.

dfed:
I usually don’t make releases without functional changes. I’ll include this change in the next functional release. Thank you for your contribution!

PR #340, opened 2025-03-14
Code cleanup, Remove "internal" access modifiers from code

richardtop:
Remove "internal" access modifiers from code #339


dfed (approved):
Dig it.

PR #357, opened 2026-03-25
Add support for multiple Valets within explicitly set shared groups

antons:
Following up on #348, this extends #297 to explicitly set shared group Valets.

### DIFF
--- a/Sources/Valet/Internal/Service.swift
+++ b/Sources/Valet/Internal/Service.swift
@@ -20,8 +20,8 @@ import Foundation
 enum Service: CustomStringConvertible, Equatable, Sendable {
     case standard(Identifier, Configuration)
     case sharedGroup(SharedGroupIdentifier, Identifier?, Configuration)
-    case standardOverride(service: Identifier, Configuration)
-    case sharedGroupOverride(service: SharedGroupIdentifier, Configuration)
+    case standardOverride(Identifier, Configuration)
+    case sharedGroupOverride(SharedGroupIdentifier, Identifier?, Configuration)
 
     // MARK: Equatable
     
@@ -49,10 +49,6 @@ enum Service: CustomStringConvertible, Equatable, Sendable {
         }
     }
 
-    static func sharedGroup(with configuration: Configuration, explicitlySetIdentifier identifier: Identifier, accessibilityDescription: String) -> String {
-        "VAL_\(configuration.description)_initWithSharedAccessGroupIdentifier:accessibility:_\(identifier)_\(accessibilityDescription)"
-    }
-
     // MARK: Internal Methods
     
     func generateBaseQuery() -> [String : AnyHashable] {
@@ -77,8 +73,8 @@ enum Service: CustomStringConvertible, Equatable, Sendable {
         case let .standardOverride(_, desiredConfiguration):
             configuration = desiredConfiguration
 
-        case let .sharedGroupOverride(identifier, desiredConfiguration):
-            baseQuery[kSecAttrAccessGroup as String] = identifier.description
+        case let .sharedGroupOverride(groupIdentifier, _, desiredConfiguration):
+            baseQuery[kSecAttrAccessGroup as String] = groupIdentifier.description
             configuration = desiredConfiguration
         }
         
@@ -110,8 +106,8 @@ enum Service: CustomStringConvertible, Equatable, Sendable {
             service = Service.sharedGroup(with: configuration, groupIdentifier: groupIdentifier, identifier: identifier, accessibilityDescription: configuration.accessibility.description)
         case let .standardOverride(identifier, _):
             service = identifier.description
-        case let .sharedGroupOverride(identifier, _):
-            service = identifier.groupIdentifier
+        case let .sharedGroupOverride(groupIdentifier, identifier, _):
+            service = identifier?.description ?? groupIdentifier.groupIdentifier
         }
 
         switch self {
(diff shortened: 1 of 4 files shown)


Sources/Valet/Valet.swift
```diff
@@ -87,8 +87,8 @@ public final class Valet: NSObject, Sendable {
     /// - Returns: A Valet that reads/writes keychain elements that can be shared across applications written by the same development team.
     /// - Warning: Using an explicitly set kSecAttrService bypasses this project’s guarantee that one Valet type will not have access to one another type’s key:value pairs. To maintain this guarantee, ensure that each Valet’s identifier is globally unique.
     /// - SeeAlso: https://github.com/square/Valet/issues/140
-    public class func sharedGroupValet(withExplicitlySet identifier: SharedGroupIdentifier, accessibility: Accessibility) -> Valet {
-        findOrCreate(explicitlySet: identifier, configuration: .valet(accessibility))
+    public class func sharedGroupValet(withExplicitlySet groupIdentifier: SharedGroupIdentifier, identifier: Identifier? = nil, accessibility: Accessibility) -> Valet {
+        findOrCreate(explicitlySet: groupIdentifier, identifier: identifier, configuration: .valet(accessibility))
     }
```
dfed, line 93: need to update documentation comment given the new parameters

Sources/Valet/Valet.swift
```diff
@@ -98,8 +98,8 @@ public final class Valet: NSObject, Sendable {
     /// - Returns: A Valet (synchronized with iCloud) that reads/writes keychain elements that can be shared across applications written by the same development team.
     /// - Warning: Using an explicitly set kSecAttrService bypasses this project’s guarantee that one Valet type will not have access to one another type’s key:value pairs. To maintain this guarantee, ensure that each Valet’s identifier is globally unique.
     /// - SeeAlso: https://github.com/square/Valet/issues/140
-    public class func iCloudSharedGroupValet(withExplicitlySet identifier: SharedGroupIdentifier, accessibility: CloudAccessibility) -> Valet {
-        findOrCreate(explicitlySet: identifier, configuration: .iCloud(accessibility))
+    public class func iCloudSharedGroupValet(withExplicitlySet groupIdentifier: SharedGroupIdentifier, identifier: Identifier? = nil, accessibility: CloudAccessibility) -> Valet {
```
dfed, line 103: same here

Sources/Valet/Internal/Service.swift
```diff
@@ -49,10 +49,6 @@ enum Service: CustomStringConvertible, Equatable, Sendable {
         }
     }
 
-    static func sharedGroup(with configuration: Configuration, explicitlySetIdentifier identifier: Identifier, accessibilityDescription: String) -> String {
-        "VAL_\(configuration.description)_initWithSharedAccessGroupIdentifier:accessibility:_\(identifier)_\(accessibilityDescription)"
-    }
```
dfed, line 54: previously it was possible to have an explicitly set identifier that was not a shared group identifier. it looks like we've removed that ability? is this a breaking change?

Sources/Valet/Valet.swift
```diff
@@ -805,7 +843,7 @@ extension Valet {
 
     class func permutations(withExplictlySet identifier: SharedGroupIdentifier) -> [Valet] {
         Accessibility.allCases.map { accessibility in
-            .sharedGroupValet(withExplicitlySet: identifier, accessibility: accessibility)
+                .sharedGroupValet(withExplicitlySet: identifier, identifier: nil,  accessibility: accessibility)
```
dfed: why the whitespace change?

dfed (changes requested):
I'm not entirely sure I understand what problem is being solved here. Our [contributing guidelines](https://github.com/square/Valet/blob/main/Contributing.md) suggest creating an issue first so we can align before we're at the code-review stage.

Since we're already at the PR stage, the PR description itself needs to be much more fleshed out. What is the motivation? What change is being proposed? Is this change backwards compatible? What should someone adopting this proposed API take into consideration?

Given the potential for foot guns that this proposed API has, I think we likely need to improve the in-code documentation as well. But we can do that once we're aligned on what we're solving.

Sources/Valet/Internal/Service.swift
antons: This was not a public API. As far as I can see, this was effectively dead code. It was only referenced by the migration path from v3 Valets, and shared group Valets with explicitly set identifiers did not exist in v3.

Sources/Valet/Valet.swift
antons: Fixed.

Sources/Valet/Valet.swift
antons: Fixed.

Sources/Valet/Valet.swift
antons: Fixed.

antons:
@dfed I’m sorry about the terse description. At this point it’s been more than 6 months since I implemented this, at the same time as #348 (which took many months to merge due to CLA issues), so I had to dive into this again:

#297 skipped the (at that time macOS-only) explicitly set identifier API. Shared group Valets continued to use `groupIdentifier.groupIdentifier` as the `kSecAttrService`. `groupIdentifier.groupIdentifier` is the reverse-DNS-formatted group ID, and on macOS this would be shown to users in Keychain Access prompts.

This PR adds the ability to provide an explicitly set `kSecAttrService` for shared group Valets. Before this change, `net.ia.account` would be shown to our users for sign-in data that is shared across iA Writer and iA Presenter. With this change, we can specify “iA Account”.

This API shares all the caveats of the already-existing explicitly set identifier API.

